PR #9, opened 2024-02-11
Svg images

csagataj2:
This PR replaces most of the png icons with their vector image counterpart. This method needs inkscape for compilation, so I modified the workflow too.


main.tex
```diff
@@ -198,13 +200,13 @@ \subsection*{Town actions}
 \end{itemize}
 
 \subsection*{Morale actions}
-Players can gain or lose morale through various game effects. When a player gains morale, place a positive morale token \includegraphics[scale=0.75]{images/positive.PNG} near your play area. A player may only have one such token. If they would gain a second token, they may immediately spend the first one before gaining the second. A positive morale token may be spent to perform any of the following actions at any time:
+Players can gain or lose morale through various game effects. When a player gains morale, place a positive morale token  \includesvg[height=10px]{svgs/positive.svg} near your play area. A player may only have one such token. If they would gain a second token, they may immediately spend the first one before gaining the second. A positive morale token may be spent to perform any of the following actions at any time:
```
qwrtln, line 203:
```suggestion
Players can gain or lose morale through various game effects. When a player gains morale, place a positive morale token \includesvg[height=10px]{svgs/positive.svg} near your play area. A player may only have one such token. If they would gain a second token, they may immediately spend the first one before gaining the second. A positive morale token may be spent to perform any of the following actions at any time:
```
Extra space

main.tex
```diff
@@ -293,8 +295,8 @@ \subsection*{\hypertarget{Ability}{Ability and statistic cards}}
...
 
 \end{wrapfigure}
-Every ability and statistic card has a basic effect and a stronger expert effect, which is shown below the basic effect. Whenever a player plays an ability or statistic card, they must choose which effect they are using. The number of expert \includegraphics[scale=0.5]{images/expert.PNG} effects a player can use each round is limited by their main hero’s \hyperlink{Level}{level}.\\[6pt]
-\textbf{Important}: Certain cards are limited to the Necropolis faction \includegraphics[scale=1]{images/necro.PNG}. When a non-Necropolis player draws one, they may either discard it and draw a new card as a replacement or gain it as normal. Non-Necropolis players cannot use faction specific cards from their hand in any way besides discarding them.
+Every ability and statistic card has a basic effect and a stronger expert effect, which is shown below the basic effect. Whenever a player plays an ability or statistic card, they must choose which effect they are using. The number of expert \includesvg[height=10px]{svgs/expert.svg} effects a player can use each round is limited by their main hero’s \hyperlink{Level}{level}.\\[6pt]
```
qwrtln, line 298: Can you delete those PNG files? I guess we won't be needing them any more.

main.tex
```diff
@@ -497,21 +499,21 @@ \subsection*{\hypertarget{Combatterminology}{Combat terminology}}
...
 \clearpage
 \subsection*{\hypertarget{Walls}{Defending a town with a citadel}}
-When a town with a citadel is attacked, the defender adds the wall and gate obstacles in any order to the middle row of the combat board after placing their units. The gate card is not an obstacle to the defending player. The wall and gate cards can be destroyed by any adjacent \includegraphics[scale=0.25]{images/ground_icon.PNG} or \includegraphics[scale=0.25]{images/flying_icon.PNG} unit's attack. Such attacks are always successful, do not roll the combat die when performing them.\\[6pt]
+When a town with a citadel is attacked, the defender adds the wall and gate obstacles in any order to the middle row of the combat board after placing their units. The gate card is not an obstacle to the defending player. The wall and gate cards can be destroyed by any adjacent \includesvg[height=10px]{svgs/unit_ground.svg} or \includesvg[height=10px]{svgs/unit_flying.svg} unit's attack. Such attacks are always successful, do not roll the combat die when performing them.\\[6pt]
```
qwrtln, line 516: It's been a while since I've used LaTeX myself, but I wonder if it's possible to use a variable instead of writing `10px` everywhere? Say somebody decides to use 11px or 9px in the future, this change would be a lot less painful.

csagataj2: It can be done, thanks for the suggestion.

main.tex
csagataj2: Sure thing

qwrtln:
Looks very neat! What do you think about unifying the "images" section? We could create a single `assets` directory, and hold separate directories `images` and `svgs` within.

main.tex
qwrtln: I guess there are a number of hardcoded values like that now. We could help @Heegu-sama identify them and put them in the beginning of the document as variables.

Heegu-sama: Yes, the most common of these is the constant [6pt] when changing to a new paragraph. The default paragraph spacing should be defined as that.

Heegu-sama:
Would you happen to have a clear set of instructions for setting up TexStudio to convert svg files when building the pdf? I had some time yesterday to dabble into this, but I'm confused that I couldn't find a concise, complete list of steps to set it up properly :blush:

I've installed the svg package using MiKTeX and inkscape. I also added the command "--shell-escape" to pdflatex.exe's options.

csagataj2:
> Would you happen to have a clear set of instructions for setting up TexStudio to convert svg files when building the pdf? I had some time yesterday to dabble into this, but I'm confused that I couldn't find a concise, complete list of steps to set it up properly 😊
> 
> I've installed the svg package using MiKTeX and inkscape. I also added the command "--shell-escape" to pdflatex.exe's options.

I think you are missing inkscape from the Path variable.

qwrtln:
> Would you happen to have a clear set of instructions for setting up TexStudio to convert svg files when building the pdf? I had some time yesterday to dabble into this, but I'm confused that I couldn't find a concise, complete list of steps to set it up properly 😊
> 
> I've installed the svg package using MiKTeX and inkscape. I also added the command "--shell-escape" to pdflatex.exe's options.

You shouldn't need any special setup. With `--shell-escape` option, it should be enough just to install `inkscape` on your machine. At least it did the trick for me, I'm not sure if Windows has the same setup.

csagataj2:
> > Would you happen to have a clear set of instructions for setting up TexStudio to convert svg files when building the pdf? I had some time yesterday to dabble into this, but I'm confused that I couldn't find a concise, complete list of steps to set it up properly 😊
> > I've installed the svg package using MiKTeX and inkscape. I also added the command "--shell-escape" to pdflatex.exe's options.
> 
> You shouldn't need any special setup. With `--shell-escape` option, it should be enough just to install `inkscape` on your machine. At least it did the trick for me, I'm not sure if Windows has the same setup.

I tested this on Windows and I needed the extra Path edit.

Heegu-sama:
> > > Would you happen to have a clear set of instructions for setting up TexStudio to convert svg files when building the pdf? I had some time yesterday to dabble into this, but I'm confused that I couldn't find a concise, complete list of steps to set it up properly 😊
> > > I've installed the svg package using MiKTeX and inkscape. I also added the command "--shell-escape" to pdflatex.exe's options.
> > 
> > 
> > You shouldn't need any special setup. With `--shell-escape` option, it should be enough just to install `inkscape` on your machine. At least it did the trick for me, I'm not sure if Windows has the same setup.
> 
> I tested this on Windows and I needed the extra Path edit.

Right, if I recall correctly the final error it produced last night was indeed about the path. I'll check that out today.

Heegu-sama (approved):
Excellent stuff. There are still some additional svg files that could be ripped, but merging this version is already long overdue so I'll do it now. Page 19 (unit card anatomy) broke its wrapping around the image in this build but I'll fiddle with it later. This is another image that I'll probably recreate from scratch later anyway.

PR #13, opened 2024-02-13
Unify assets

qwrtln:
This PR does the following:
1. Puts images and SVGs in a single `assets` directory with separate subdirectories
2. Introduces variables for paths to the above
3. Unifies files structure and removes leftover PNGs


Heegu-sama (approved):
Looking good. Builds identically.

PR #55, opened 2024-02-27
Movement token to svg

csagataj2:


sections/player_turns.tex
```diff
@@ -25,7 +25,7 @@ \subsection*{Movement Actions}
 \end{itemize}
 
 \begin{center}
-  \includegraphics[width=0.6\linewidth]{\images/movement_tokens.png}\\
+  \includesvg[width=0.6\linewidth]{\images/movement_tokens.svg}\\
```
qwrtln, line 28:
```suggestion
  \includesvg[width=0.6\linewidth]{\svgs/movement_tokens.svg}\\
```

qwrtln (changes requested):
Shouldn't it go to the `glyphs` folder instead? Also, please get rid of the old image.

qwrtln:
> > Shouldn't it go to the `glyphs` folder instead? Also, please get rid of the old image.
> 
> I think glyphs were named glyphs because they were extracted from the fonts glyphs section. But I can move it there.

OK, it's fine then. Remove the old image and that's it.

qwrtln (approved):

Heegu-sama (approved):

PR #62, opened 2024-03-02
Fixed some capitalization inconsistencies

RickardNi:
I tried to fix problems with inconsistencies, mainly those regarding capitalization.

There's been a discussion about this in the BGG thread and I welcome further discussions about details. Many of the changes are straight-forward, but some of them can raise questions, but there's a motivation for each and every change from my perspective, so I welcome questions and suggestions.

Overall, I think we need to decide what "vague" terms should be _game terms_ and not just _general terms_. For instance, terms like _Card_, _Deck_, and _Turn_ were mostly capitalized, and thus, I tried to be consistent with that (though, not everything is complete yet, but this is a good start). Other terms like _player_ and _hand_ are not capitalized since they are not considered _game terms_.

I have been quite generous with the amount of capitalization, but that's mainly to keep everything intact and consistent with how it was written already. If the consensus is that it's too much, that's great, but then we should decide that they are not capitalize and have it consistent everywhere.

I also want to raise another option that exist for some terms like _token_. They are capitalized only when it is clearly a game term, like Positive Morale Token, but not capitalized when they are spoken in general, like:
> If they would gain a second token...

This is an apporach that makes sense and could be applied to terms like _card_ and _deck_ as well, but then we need to decide upon that, because right now (before this PR) it is scattered and inconsistent.


qwrtln (approved):
Thank you for this, this improves the quality, and makes thing look more professional 💪🏻

Heegu-sama (approved):

PR #63, opened 2024-03-02
Set art opacity to 50% and reduce oversized images

qwrtln:
This change reduces PDF size from 51 to 42MB.

A slight change to gremlin and gorgone:
image
image


RickardNi:
Looks fine, but I have some questions:

1. Is the opacity added for style, or to further reduce the pdf size?
2. Is there a long-term goal to reduce pdf size as much as possible?
3. Would it be a possibility to have a high-quality version and a size-reduced version, or would that be a pain to handle?

qwrtln:
> Looks fine, but I have some questions:
> 
>     1. Is the opacity added for style, or to further reduce the pdf size?
> 
>     2. Is there a long-term goal to reduce pdf size as much as possible?
> 
>     3. Would it be a possibility to have a high-quality version and a size-reduced version, or would that be a pain to handle?

1. Just style, it doesn't change images' size significantly. Look at the diff in size shown by GitHub.
2. Possibly, as IMO over 50MB is way too much for a file of this size, even with this much art.
3. No point. We can restore them from git history, shall we need/want to.

RickardNi (approved):

Heegu-sama (approved):

PR #67, opened 2024-03-04
Fix Sandro example

qwrtln:
Fixed some mistakes, made the example consistent with available assets.

sandro


qwrtln:
sandro-21

RickardNi (approved):
I don't fully understand the rules yet, so I can't speak to any mistakes made there, but it looks good.

We should decide whether "alternative" should be a game term or not to keep it consistent. Right now, the original rulebook only mentions it twice and it does not exist in the index.

qwrtln:
> I don't fully understand the rules yet, so I can't speak to any mistakes made there, but it looks good.
> 
> We should decide whether "alternative" should be a game term or not to keep it consistent. Right now, the original rulebook only mentions it twice and it does not exist in the index.

Yes, it's the official term, straight from the original rule book. You can also see it in this rule book on page 16.

qwrtln:
Originally, the example mentioned empowering Magic Arrow by 3. It's not possible, as it can be empowered only by 2 :D Also, there's no other spell with such empowerment. It can be either 1, 2 or 4.
Additionally, capturing gold mine was explained, while I included a tile with stones mine 🤦🏻

qwrtln:
...and it's not capitalised on page 16. I changed it here for consistency.

RickardNi:
> ...and it's not capitalised on page 15. I changed it here for consistency.

Yeah, exactly, that's what I meant by "game term". I know it existed in the original rulebook, but we have "loosely" decided that game terms should be capitalized, and other terms should not. Thus, I was wondering if we should decide "alternative" to be such a "game term" or not.

As you changed it (I didn't dare to do it since it was debatable), we now have full consistency for the word in the new rulebook, and thus it is not a game term.

Examples of "game terms" in the new rulebook: Deck, Hero, Scenario, Round

Examples of non-game terms: player, hand, defeated, competitive, expansion

TLDR; I agree with you.

PR #102, opened 2024-03-19
Add components list document

qwrtln:
This will be built just like the rulebook is. It can be distributed with it too.

See preview of the document: https://cdn.discordapp.com/attachments/1206576497331019786/1220480794796101692/components_list.pdf?ex=660f1850&is=65fca350&hm=04a93deceb43147807210c74fc5ba9c9180cc9375fb51a720bb968f01d5be41f&


RickardNi:
This is a nice addition. Overall it looks good!

Minor things:

- "This document aims to list comprehensively the contents of all the board game boxes."
I might be wrong here, and English is not my native language, but to me, it seems like the words 'list' and 'comprehensively' should switch?

- I don't understand the logic of why some "headlines" are bold and some are not. Like "Combat Board", "Round Tracker" and "Cards" are all bold, but "Unit Cards" are not. Overall I think it would be nice with some space below these headlines, and I would like to see the "top headlines", where I include "Unit Cards" to all be bold at least. Especially since we have a second level headline like "Castle" and "Necropolis".

- Some second level headlines like "Alamar" and "Mutare" are very tight and makes it harder to differeniate. I think this is because we want to fit everything in two pages, but personally, I rather extend it to a new page or decrease the font size.

- For tokens that are further specified, like "Build tokens", I would prefer some character/token before the list. A hyphen or something perhaps.

- Minor capitalization inconsistencies, but that might also depend on how the components actually are written. I will adjust these later when I receive my copy.

qwrtln:
> This is a nice addition. Overall it looks good!
> 
> Minor things:
> 
>     * "This document aims to list comprehensively the contents of all the board game boxes."
>       I might be wrong here, and English is not my native language, but to me, it seems like the words 'list' and 'comprehensively' should switch?
> 
>     * I don't understand the logic of why some "headlines" are bold and some are not. Like "Combat Board", "Round Tracker" and "Cards" are all bold, but "Unit Cards" are not. Overall I think it would be nice with some space below these headlines, and I would like to see the "top headlines", where I include "Unit Cards" to all be bold at least. Especially since we have a second level headline like "Castle" and "Necropolis".
> 
>     * Some second level headlines like "Alamar" and "Mutare" are very tight and makes it harder to differeniate. I think this is because we want to fit everything in two pages, but personally, I rather extend it to a new page or decrease the font size.
> 
>     * For tokens that are further specified, like "Build tokens", I would prefer some character/token before the list. A hyphen or something perhaps.
> 
>     * Minor capitalization inconsistencies, but that might also depend on how the components actually are written. I will adjust these later when I receive my copy.

Great comments, thanks!

1. Fixed
2. The logic is the following: every "type" of component, be it a booklet, board, die, token, miniature or card, is <ins>**bold and underlined**</ins>, while every subset of those (like a specific type of cards) is <ins>underlined</ins> only. I agree, it makes the whole list look "uneven", as cards are a dominant component of this game. I'm open to ideas regarding this.
3. I'll try experimenting with this and see what I can come up with.
4. You mean those listed as "blue, purple, green"? I actually thought about removing this specification. What do you think?
5. Yeah, I haven't paid much attention to that. I just copied the source material, formatted it, sorted, and left the original spelling as it was. Any help with this is greatly appreciatd.

qwrtln:
@RickardNi All fixed, the PDF link has been updated. Pasting it here too: https://cdn.discordapp.com/attachments/1206576497331019786/1220480794796101692/components_list.pdf?ex=660f1850&is=65fca350&hm=04a93deceb43147807210c74fc5ba9c9180cc9375fb51a720bb968f01d5be41f&

RickardNi:
Yeah I think your latest changes were a huge improvement. Everything looks much cleaner and neat now.

Great job! :)

RickardNi (approved):

Heegu-sama (approved):

PR #128, opened 2024-04-05
[DO NOT MERGE] Add support for multiple languages

qwrtln:


.github/workflows/build-rule-book.yaml
```diff
@@ -16,6 +16,12 @@ on:
 jobs:
   build_rule_book:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
evilmint, line 20: is this to catch all errors in all languages just in case?

README.md
```diff
@@ -59,11 +59,11 @@ To work on the document on your machine, you need the following:
 To build the document, either run this in the command line:
 
 ```bash
-latexmk -pdf -silent -shell-escape "main"
+latexmk -pdf -silent -shell-escape "main_en"
```
evilmint, line 62: would it be worth adding a comment that the en part might be substituted for non-techy people?

README.md
```diff
@@ -4,8 +4,8 @@ Please see the original thread on [BoardGameGeek](https://boardgamegeek.com/thre
 
 This repository hosts **three** documents. Click to download the most recent builds:
 
-- [📜 **Rewritten Rule Book**](https://github.com/Heegu-sama/Homm3BG/raw/build/main.pdf/PDF/main.pdf)
-- [🖨️ **Rewritten Rule Book - printable version️️**](https://github.com/Heegu-sama/Homm3BG/raw/build/printable.pdf/PRINTABLE/printable.pdf)
+- [📜 **Rewritten Rule Book**](https://github.com/Heegu-sama/Homm3BG/raw/build/main_en.pdf/PDF/main_en.pdf)
+- [🖨️ **Rewritten Rule Book - printable version️️**](https://github.com/Heegu-sama/Homm3BG/raw/build/printable.p_en.pdf/PRINTABLE/printable_en.pdf)
 - [📋 **Comprehensive Components List**](https://github.com/Heegu-sama/Homm3BG/raw/build/components_list.pdf/COMPONENTS_LIST/components_list.pdf)
```
evilmint, line 9: do we want to include links to other translations' builds here as well?

metadata.tex
```diff
@@ -198,19 +192,26 @@
...
+  \en{The Rewrite}
+  \pl{}
+}
 
 \author{Hermanni "Heegu" Karppela}
 \maketitle
 
 \begin{center}
-  Version 1.0
+  \en{Version 1.0}
```
evilmint, line 204: wdyt about moving the version literal to a constant in case there are far more languages?

sections/combat.tex
```diff
@@ -1,93 +1,159 @@
...
-Then, place any \includesvg[height=10px]{\svgs/unit_ground.svg} or \includesvg[height=10px]{\svgs/unit_flying.svg} Units in the Frontline.
-If there's not enough room to place a Unit in its correct line, place them in the other one.
-Units must be placed in \textbf{descending} Initiative\index{Initiative} order.
-If there's a tie, place higher tier Units first.
-If there's still a tie, the players decide the order.
+  \en{
+    \item Choose one of the Combat Board's sides as your own.
+      Place up to 5 of your Unit cards freely onto the Back and Frontlines of that side.
+  }
+  \pl{\item}
```
evilmint, line 40: I'm thinking about a quick way to find untranslated strings. If we search by \pl{} we won't find all occurrences because of this case for example. is it possible and worth having an empty \language{} element always, or if not maybe replacing the whole string with a literal like TTR (to translate)? thinking about looking for untranslated strings which there will be many when a new language starts off

.github/workflows/build-rule-book.yaml
qwrtln: If set to true (the default), when a single job fails, all the other jobs are cancelled. We don't want that. We want all of them to complete.

README.md
qwrtln: Yes we do, but we don't want to store them in the repo. With two languages, it will be over 300 MB of binary storage to pull after merging every PR. We don't want that. We'll have to find a way to store it externally.

sections/combat.tex
qwrtln: Yes, good idea. A PR with this is welcome. I put those tags with `\item`, because LaTeX returns an error, if there's `\begin{itemize}` and no `\item` inside.

metadata.tex
qwrtln: Interesting idea, but how do you imagine this exactly? Sounds a little like you're thinking JavaScript 😆

README.md
qwrtln: Yes, definitely. I'm thinking of revamping the readme totally, but in a separate PR.

metadata.tex
qwrtln: Or, we ditch "Version" and simply do `v1.0`

PR #159, opened 2024-04-13
Linkify more links in printable

DarthGandalf:
Why is there such list at all, instead of replacing all hyperlink with pagelink?


qwrtln:
Firstly, if you'd like to use all the hyperlinks, it's better to get rid of the script altogether. Secondly, did you try building this? Appending page number to every single hyperlink produced so much text, that the layout kept breaking.

DarthGandalf:
Yes, it builds and looks correct. Layout broke on which page for you?

qwrtln:
No page in particular. The build was automatic, so it kept breaking from time to time. Fixing it by hand was a nightmare. If you let it running like that, it will keep breaking until somebody notices it. We noticed that limiting those hyperlinks gives consistent results.
You think it's more readable with all hyperlinks appended with page numbers?

DarthGandalf:
>  You think it's more readable with all hyperlinks appended with page numbers?

For the printed version, yes. Otherwise, I need to mentally figure out what the topic could be, and find it either in TOC, or in index.

qwrtln:
Let's park this for now.

PR #163, opened 2024-04-14
Mention 4p game in intro

### LINKED ISSUE #132: Proposal for specifying number of players with expansions
Currently the introduction mentions that the core game box set allows for 1-3 players.
I was thinking if it's worth adding information that with each expansion brought into the game, the game allows for 1 more player so that people know it's a single expansion that raises the maximum player count possible?

DarthGandalf:
Also visually split thematic fluff from the explanation of what the game technically is

Close #132

### DIFF
--- a/sections/introduction.tex
+++ b/sections/introduction.tex
@@ -3,8 +3,9 @@
 
 \bigbreak
 
-\hypertarget{Heroes of Might and Magic III}{\textbf{Heroes of Might and Magic III: The Board Game}} is a tactical strategy RPG board game for 1-3 players using the core box set.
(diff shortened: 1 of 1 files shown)


qwrtln:
This is not entirely correct. Firstly, the core box set allows up to 3 players. Secondly, with custom scenarios the only limiting factor is the number of factions, so currently you can have as many as 7 players. There were people on BGG saying they played a successful 6-player scenario. Thirdly, the fourth player is mentioned on the next page.

DarthGandalf:
Should it say instead "more with expansions" without specifying the number 4?

If no, probably the issue linked should be closed.

Either way, I'd still visually separate the fluff

PR #172, opened 2024-04-15
Make resource minitiatures to be in same section

DarthGandalf:


qwrtln (changes requested):
Please update the respective token count.

DarthGandalf:
Fixed

qwrtln (approved):

PR #180, opened 2024-04-17
Update plastic coin (1) count from 11 -> 10

evilmint:
According to others, myself included - there are 10 Plastic Coins (1)


qwrtln (changes requested):
Please update the plastic token count in the header.

qwrtln (approved):

PR #201, opened 2024-04-20
Add Difficulty section and a table of optional rules

DarthGandalf:
Because the table is so big, add another page, and move some graphics. Because cerberi is not used anymore, put it back to Units page, where it's more appropriate than the magic arrow.

Move difficulty table to the back cover

both-27
both-37

Ignore the left half of the following, that gets moved to page 41
both-40
both-51


qwrtln:
Please add `figure*` to the literal expressions in line 6 of `po4a.cfg`.

sections/other_rules.tex
```diff
@@ -46,14 +46,21 @@ \subsection*{\hypertarget{Difficulty}{Difficulty}}
...
-\hfill\includegraphics[width=0.8\linewidth]{\art/cerberi.png}
+\includegraphics[width=\linewidth]{\art/sharpshooter.jpg}
 
-\end{multicols}
+\clearpage
 
 \begin{figure*}[!hb]
   \centering
-  \includegraphics[width=\linewidth]{\tables/field_difficulty.png}
+  \includegraphics[width=\linewidth]{\tables/opt_rules.png}
```
qwrtln, line 60:
```suggestion
  \includegraphics[width=\linewidth]{\tables/optional_rules.png}
```
Make it explicit.

qwrtln: BTW, the image is not properly cropped. Can you get rid of the areas beyond the yellow frame?

sections/other_rules.tex
```diff
@@ -46,14 +46,21 @@ \subsection*{\hypertarget{Difficulty}{Difficulty}}
...
 \end{figure*}
+
+\begin{scaledfigure}[blanker]
+  \centering
+  \includegraphics[width=\linewidth, height=\myspace, keepaspectratio]{\art/magic_arrow.png}
```
qwrtln, line 65: Magic Arrow doesn't fit here. Can you try something else? Actually, a full framed graphic would fit better, not a transparent one. If you find an appropriate graphic in any of the original rule books, I can rip it for you. It must be wide but not tall.

sections/back_cover.tex
```diff
@@ -5,3 +5,9 @@
     \includegraphics[height=\paperheight, keepaspectratio]{\layout/back.jpg}
   };
 \end{tikzpicture}
+
+\begin{figure*}[!hb]
+  \hypertarget{Difficulty Table}{}
+  \centering
+  \includegraphics[width=\linewidth]{\tables/field_difficulty.png}
+\end{figure*}
```
qwrtln, line 13:
```suggestion
\begin{figure}[!h]
  \hypertarget{Difficulty Table}{}
  \centering
  \includegraphics[width=\linewidth]{\tables/field_difficulty.png}
\end{figure}
```
I think most of those configs are unnecessary. With this only it should work, and it won't confuse `po4a`?

sections/back_cover.tex
```diff
@@ -5,3 +5,9 @@
     \includegraphics[height=\paperheight, keepaspectratio]{\layout/back.jpg}
   };
 \end{tikzpicture}
+
+\begin{figure*}[!hb]
+  \hypertarget{Difficulty Table}{}
+  \centering
+  \includegraphics[width=\linewidth]{\tables/field_difficulty.png}
```
qwrtln, line 12: Please crop out the drop shadow of this image. It's no longer necessary on the back cover.

sections/combat.tex
```diff
@@ -21,7 +21,7 @@ \subsection*{\hypertarget{Combatsetup}{Combat Setup}}
 \begin{itemize}
   \item Choose one of the Combat Board's sides as your own.
     Place up to 5 of your Unit cards freely onto the Back and Frontlines of that side.
-  \item Check the \hyperlink{Difficulty}{\textbf{Difficulty Table}} and draw the corresponding number of Neutral Unit cards from their Decks.
+  \item Check the \hyperlink{Difficulty Table}{\textbf{Difficulty Table}} and draw the corresponding number of Neutral Unit cards from their Decks.
```
qwrtln, line 24: In the printable build, there will be page 52 here 😆 You think that's OK?

qwrtln (changes requested):
Please make the graphical adjustments.

sections/combat.tex
DarthGandalf: I'm open for suggestions

sections/other_rules.tex
DarthGandalf: Couldn't find anything small enough. Will just remove it

qwrtln:
The sharpshooter doesn't look good, because it's too far away from the text. Can you find something taller?

sections/combat.tex
qwrtln: I mean, the printable version will say explicitly (p. 52), but there is no such page. I guess we should exclude this from the Python script.

DarthGandalf:
I moved Difficulty to a brand new section, before Scenario End. PTAL

sections/difficulty.tex
```diff
@@ -0,0 +1,29 @@
+% !TeX spellcheck = en_US
+\addsection{Difficulty}{\skills/water_magic.png}
```
qwrtln, line 2: I don't like water magic here. I suggest one of the following: stone skin, haste, fire magic. In my head, I have a slight preference for stone skin, but please check them and choose what looks best.

qwrtln: It's fine.

qwrtln (approved):

Heegu-sama (approved):

sections/combat.tex
DarthGandalf: PTAL. In future we should choose which link should add (p. N) and which should not, and remove the script.

metadata.tex
```diff
@@ -161,6 +161,8 @@
 % Commands to be used for automation generating printable version
 \newcommand{\pagetarget}[2]{\label{#1}\hypertarget{#1}{#2}}
 \newcommand{\pagelink}[2]{\hyperlink{#1}{#2} (\pageshorthand \pageref{#1})}
+% Like a \hyperlink but doesn't append page number in printable version
+\newcommand{\hyperlinkwopage}[2]{\hyperlink{#1}{#2}}
```
qwrtln, line 165: There's no need for that. Since you made it a separate hyperlink from just `Difficulty`, it won't have page number appended by default.

qwrtln, line 165:
```suggestion
```

DarthGandalf: Are you sure?
Выделение_524

PR #206, opened 2024-04-21
Make build script platform independent

qwrtln:


build.sh
```diff
@@ -1,8 +1,15 @@
 #!/usr/bin/env bash
 
+unameOut="$(uname -s)"
+case "${unameOut}" in
+    Linux*)     command=xdg-open;;
+    Darwin*)    command=open;;
+esac
+echo ${command}
+
 LANGUAGE=$1
```
evilmint, line 10: exit 1 if arg wasn't found? else the process carries on

build.sh
```diff
@@ -1,8 +1,15 @@
 #!/usr/bin/env bash
 
+unameOut="$(uname -s)"
+case "${unameOut}" in
+    Linux*)     command=xdg-open;;
+    Darwin*)    command=open;;
+esac
+echo ${command}
```
evilmint, line 8: I guess was needed for debug only

evilmint (approved):
works!

DarthGandalf:
I'm fine with having this script here, but it's useless to me due to `rm` part of it. Every rerun of the script is very slow, and makes iteration on the text slow. Why is rm needed?

If you want me to use it, remove rm; or I can just stick to using latexmk directly.

build.sh
qwrtln: It's fine. Fails exactly where it should.

qwrtln:
> I'm fine with having this script here, but it's useless to me due to `rm` part of it. Every rerun of the script is very slow, and makes iteration on the text slow. Why is rm needed?
> 
> If you want me to use it, remove rm; or I can just stick to using latexmk directly.

If all you need is `latexmk`, you don't need a script anyway.

PR #217, opened 2024-04-22
Fix pl index

qwrtln:


.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -108,10 +108,16 @@ jobs:
         uses: xu-cheng/latex-action@v3
         with:
           extra_system_packages: "inkscape"
+          pre_compile: echo "icu_locale \"${{ matrix.language }}\"" >> index_style.ist && upmendex -s index_style main_${{ matrix.language }} && sed -i 's@\(\\noindent\\textbf{\)\(.\)@\\noindent\\textbf{\2}@g' main_${{ matrix.language }}.ind
           latexmk_shell_escape: true
-          post_compile: "makeindex main_${{ matrix.language }} -s index_style.ist && pdflatex --shell-escape main_${{ matrix.language }}.tex"
           root_file: main_${{ matrix.language }}.tex
 
+      - name: Move file
```
DarthGandalf, line 115: two different Move files?

DarthGandalf (approved):

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Yes, because the deployment steps will deploy from a directory. You move both files to the same directory

qwrtln: Argh, no copy paste error xD

PR #249, opened 2024-04-25
Delete components list

qwrtln:


qwrtln:
> Call it archive without . ?
> 
> Or just rm; the old files are still available in git history and even in the 1.1 release

Okay, I guess I got too sentimental.

DarthGandalf (approved):

PR #253, opened 2024-04-25
[ES] Combat

luischos:
combat1
combat2
combat3
combat4
combat5
combat6


qwrtln:
Page 34 has visibly more text than the original. Can you do anything about it?

luischos:
I'll try, but is not easy...

luischos:
Changed some texts, hope this could be enought

PR #257, opened 2024-04-25
Clarify rules of defending, based on Alex K's answer

DarthGandalf:


qwrtln:
Well shit, another thing that is actually written as the opposite of what they intended...

qwrtln:
What is that supposed to mean? That Defence actually replaces activation?

DarthGandalf:
No, you can move, then defend. The change is more important for ranged units: they cannot move after getting the defense token

qwrtln:
OK, it's logical, but maybe instead it's worth to add a sentence explaining that ranged units cannot move after defending? In its current form it makes me doubt if it applies to ranged units, even though I know that's not the case.

sections/combat.tex
```diff
@@ -71,7 +71,7 @@ \subsection*{\hypertarget{Combatterminology}{Combat Terminology}}
 That Unit \textbf{must skip its next Activation}. Remove the Token instead of activating it.
 If the Unit \textbf{attacked or takes any damage} before that time, \textbf{remove the Token}.
 The Token does not prevent Units from performing Retaliation Attacks.\par
-\textbf{\hypertarget{Defend}{Defend}} \includesvg[height=10px]{\svgs/defense.svg} – Units may choose to gain a Defense Token instead of attacking.
+\textbf{\hypertarget{Defend}{Defend}} \includesvg[height=10px]{\svgs/defense.svg} – Units may choose to gain a Defense Token and end the Activation instead of attacking.
```
qwrtln, line 74:
```suggestion
\textbf{\hypertarget{Defend}{Defend}} \includesvg[height=10px]{\svgs/defense.svg} – Units may choose to gain a Defense Token instead of attacking.
Ranged Units cannot move after gaining it.
```
It also makes it clear that non-ranged units can still move before defending. A lot of people have doubts around this.

DarthGandalf:
Выделение_533

This is not only about ranged units. All units end their activation if they choose to take defense token instead of attacking.

DarthGandalf:
both-30

sections/combat.tex
```diff
@@ -71,7 +71,8 @@ \subsection*{\hypertarget{Combatterminology}{Combat Terminology}}
...
+Remember that Ranged Units \includesvg[height=10px]{\svgs/unit_ranged.svg} move after attacking, therefore they won't be able to move if they choose to Defend.
```
qwrtln, line 75: I think it's unnecessary.

DarthGandalf: You just wanted me to add the clarification about ranged units :)

qwrtln: That was before you posted the picture about stacking abilities. It's unnecessary now.

DarthGandalf:
both-30

qwrtln (approved):
Thanks

Heegu-sama (approved):

PR #258, opened 2024-04-26
[FR] Player turns

Gastonia02:
player_turn_fr_page-0001
player_turn_fr_page-0002
player_turn_fr_page-0003


qwrtln (approved):
Absolutely great.

PR #260, opened 2024-04-26
[FR] heroes section

Gastonia02:
heroes_fr_page-0001
heroes_fr_page-0002
heroes_fr_page-0003


qwrtln:
Hero card is the only thing I dislike here. Can you move "Fiche de Heros" slightly to the right? Also, numbers 1-11 below it seem more condensed than in the original. Please fix it.

qwrtln (approved):
Oh, sorry, I noticed that the original is actually quite similar.

Matt-le-Fou:
> Oh, sorry, I noticed that the original is actually quite similar.

No problem. It's not centered, just idented I think.
I added a phantom to move it a bit on the right (that can be done on the original too).

I tried to fix the left column too, and the position of the griffin.

PR #269, opened 2024-04-26
[FR] map elements

Gastonia02:
map_elements_fr-22-24_page-0001
main_fr_1-23_page-0001
map_elements_fr-22-24_page-0003


qwrtln (changes requested):
You should move the node with bottom green text to the right, so that it's aligned properly.

translations/map_elements.tex/fr.po
```diff
@@ -235,6 +325,8 @@ msgid ""
 "\\includegraphics[width=\\linewidth]{\\examples/sandro_empowering_town_portal.png}\n"
 "\n"
 msgstr ""
+"\\includegraphics[width=\\linewidth]{\\examples/sandro_empowering_town_portal.png}\n"
+"\n"
```
qwrtln, line 329:
```suggestion
```

qwrtln (approved):

PR #276, opened 2024-04-27
Address developer comment

### LINKED ISSUE #274: Trading post
Выделение_538

DarthGandalf:
Fix #274

both-39


qwrtln:
Are you sure about this? My understanding was that you could do both, actually.

qwrtln:
> Are you sure about this? My understanding was that you could do both, actually.

Even so, please update the description in all map locations too.

Legrecofr:
If you read the core game rule of the trading post you can't do both it's say "or", In the Rampart rulebook buying war machines is added as a new option at the trading post. Meaning you can choose one from the 3 options.

qwrtln (approved):

Heegu-sama:
image

I re-read the original and doesn't it actually also say that you can remove only one card?

evilmint (approved):

Heegu-sama (approved):

PR #282, opened 2024-04-27
[ES] Index

luischos:
index

### DIFF
--- a/translations/index.tex/es.po
+++ b/translations/index.tex/es.po
@@ -20,4 +20,4 @@ msgstr ""
 #. type: Plain text
 #: sections/index.tex:4
 msgid "\\addsection{Index}{\\images/knowledge.png}"
-msgstr ""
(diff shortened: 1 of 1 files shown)


qwrtln (changes requested):
Remember about the header adjustment.

qwrtln:
Please show the page of the actual index. To build it, use `./make_printable es` script.

luischos:
@qwrtln ok, done both :)

qwrtln:
Huh, interesting. Is "Table of Contents" and "Index" both "Índice" in Spanish? If so, that might be confusing. Maybe there's a different name for an index page? The Index page I meant is on page 53.

luischos:
You are right. The first one could be named "Contenido", and "Índice" can go to page 54. But I don't remember where can I change it...

qwrtln (approved):

PR #298, opened 2024-04-27
[FR] scenario end

Gastonia02:
scenario_end_fr-40-1


qwrtln (changes requested):
Well, this does not look good. Is there a missing (or excessive) `vfill` somewhere?

qwrtln:
Can you restore Medusa to its original size?

PR #302, opened 2024-04-27
Fix header case and skip po4a in buld script with en

qwrtln:
Look closely at the section header text:


DarthGandalf (approved):

build.sh
```diff
@@ -13,8 +13,10 @@ else
     ENGINE=-pdf
 fi
 
+if ! [[ !${LANGUAGE}} == en ]]; then
```
DarthGandalf, line 16:
```suggestion
if [[ ${LANGUAGE}} != en ]]; then
```

qwrtln: Well that didn't work

Heegu-sama (approved):

PR #304, opened 2024-04-27
[RU] Use liberation font for headers too

DarthGandalf:
both-4
both-9


ohgodnotagain (approved):

DarthGandalf:
This is mutually exclusive with #306

qwrtln (approved):
Much better.

PR #305, opened 2024-04-27
Remove -shell-escape

DarthGandalf:
With noautomatic option of imakeidx it's not needed


qwrtln:
Are you sure? It won't work if you're doing it for the first time. `-shell-escape` runs Inkscape, too.

qwrtln (changes requested):
Try running it after `rm -rf svg-inkscape/`

DarthGandalf:
Because the PR checks didn't fail, I thought it works :( But they didn't even try to run the build

qwrtln:
> Because the PR checks didn't fail, I thought it works :( But they didn't even try to run the build

Yeah, build runs only on changed tex files and assets. Might need to rethink that approach.

PR #312, opened 2024-04-28
Find fuzzy script

qwrtln:
Demo of the script:

Screenshot from 2024-04-28 13-48-13


find_fuzzy.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
DarthGandalf, line 1: Any comment? Or entry in readme?

qwrtln: Added.

DarthGandalf (approved):

PR #278, opened 2024-04-27
Mention name of the tile

DarthGandalf:
Close #259

en-both-22

PR #326, opened 2024-04-30
Clarify exiting blocked fields

DarthGandalf:
This is what official rules say about this:

> This is a special kind of terrain that Heroes cannot enter.

Some scenarios spawn enemy heroes there, and with rewritten rules as written, they would be stuck there.

en-both-22

When combined with #278:
en-both-22


qwrtln:
Where does it say you cannot exit them?

DarthGandalf:
In the left column: "solid yellow lines cannot be passed through"

qwrtln:
> In the left column: "solid yellow lines cannot be passed through"

In that case, the not box is not accurate. You can pass them using skills or spells. Maybe this should be a regular text, not a box?

DarthGandalf:
Well, intro already states that what's written on player cards overrides the rules. How do you suggest wording this and where?

This is so corner case, that I thought the note near the place where blocked fields are mention at all would be more appropriate than interrupting the flow of text

DarthGandalf:
But it's not obvious from this "normally" whether player hero can go inside to attack the enemy hero. It's obvious from the official book - it cannot.

Heegu-sama (approved):

qwrtln:
po4a check

DarthGandalf:
po4a commit

PR #327, opened 2024-04-30
Fix empty page after Units table in printable

DarthGandalf:
en-both-27


qwrtln (changes requested):
Looks good, but please also update all the po files. It's easy to fix for you, and a nightmare to track for translators. Polish and French have it translated already.

qwrtln (approved):
Pardon me, I noticed it's untranslated.

Heegu-sama (approved):

PR #345, opened 2024-05-01
Check num of pages

qwrtln:


Heegu-sama (approved):

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -94,30 +111,25 @@ jobs:
...
 
       - name: Add pages to select hyperlinks
-        if: github.event_name != 'pull_request'
         run: |
           find ${{ matrix.sections }} -type f -execdir sed -i 's@\\hypertarget@\\pagetarget@g' '{}' +
           python .github/insert_printable_hyperlinks.py ${{ matrix.sections }}
 
       - name: Include notes page
-        if: ${{ github.event_name != 'pull_request' && env.EVEN_NUMBER_OF_PAGES == 'true' }}
+        if: ${{ env.EVEN_NUMBER_OF_PAGES == 'true' }}
```
DarthGandalf, line 123: You don't need to compare with textual 'true' like that

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -94,30 +111,25 @@ jobs:
...
       - name: Include index page
-        if: github.event_name != 'pull_request'
         run: sed -i 's@\\include{\\sections/back_cover.tex}@\\include{\\sections/index.tex}\\include{\\sections/back_cover.tex}@g' metadata.tex
 
       - name: Add QR codes
-        if: github.event_name != 'pull_request'
         run: sed -i -e '/% QR codes placeholder/{r .github/qr-codes-${{ matrix.language }}.tex' -e 'd}' metadata.tex
 
       - name: Compile printable rule book
-        if: github.event_name != 'pull_request'
```
DarthGandalf, line 120: Looks like this will make the checks longer run longer for PRs

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -126,17 +138,84 @@ jobs:
...
+      - name: Check if printable page number matches
+        if: ${{ matrix.language == 'en' }}
+        run: |
+          PAGES_DIGITAL="${{ steps.check_pages.outputs.num_of_pages }}"
+          PAGES_PRINTABLE="${{ steps.check_printable_pages.outputs.num_of_printable_pages }}"
+          echo "Number of pages in digital:   ${PAGES_DIGITAL}"
+          echo "Number of pages in printable: ${PAGES_PRINTABLE}"
+          if [[ $((${PAGES_DIGITAL}+1)) != ${PAGES_PRINTABLE} ]]; then
+            echo "Pages in digital and printable versions differ by too much. See above. Please investigate this."
+            exit 1
```
DarthGandalf, line 157: Let the artifact be downloadable even if the number of pages is wrong, otherwise it'll be very hard to debug why something overflown (on which page even); probably just move it after the upload

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -126,17 +138,84 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        language: ["pl", "es", "fr", "ru"]
+
+    steps:
+      - name: Install pdfinfo
+        run: sudo apt-get install -y poppler-utils
+
+
+      - name: Download PDFs
```
DarthGandalf, line 195: I think this need to explicitly depend on the job above, so that github doesn't run it too early

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -126,17 +138,84 @@ jobs:
...
-          external_repository: qwrtln/Homm3BG-build-artifacts
-          publish_branch: ${{ matrix.language }}
-          publish_dir: ./${{ matrix.language }}
-          force_orphan: true
+          path: main_${{ matrix.language }}
+
+      - name: Check number of pages in digital version
+        run: |
+          file main_${{ matrix.language }}/main_${{ matrix.language }}/main_${{ matrix.language }}.pdf
+          ACTUAL_NUM_OF_PAGES=$(pdfinfo main_${{ matrix.language }}/main_${{ matrix.language }}/main_${{ matrix.language }}.pdf | awk '/^Pages:/ {print $2}')
```
DarthGandalf, line 203: Why using two different mechanisms to extract the same number?

DarthGandalf (changes requested):

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Yes, no way around it.

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Ok.

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: It already depends.

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Leftover :P

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -127,10 +137,35 @@ jobs:
...
+          NUM_OF_PAGES=$(pdfinfo ${{ matrix.language }}/printable_${{ matrix.language }}.pdf | awk '/^Pages:/ {print $2}')
+          echo "NUM_OF_PRINTABLE_PAGES=${NUM_OF_PAGES}" >> "$GITHUB_OUTPUT"
+
+      - name: Check if printable page number matches
+        if: ${{ matrix.language == 'en' }}
+        run: |
+          PAGES_DIGITAL="${{ steps.check_pages.outputs.num_of_pages }}"
+          PAGES_PRINTABLE="${{ steps.check_printable_pages.outputs.num_of_printable_pages }}"
+          echo "Number of pages in digital:   ${PAGES_DIGITAL}"
+          echo "Number of pages in printable: ${PAGES_PRINTABLE}"
```
DarthGandalf, line 163: Please add a comment where +1 comes from

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -140,3 +175,43 @@ jobs:
...
           publish_dir: ./${{ matrix.language }}
           force_orphan: true
+
+  check_pages:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ["pl", "es", "fr", "ru"]
```
DarthGandalf, line 185: We have quite a few places already with this list... Will need to do some deduplication of these lists as follow-up

DarthGandalf (approved):

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Yeah, good idea.

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Sure thing.

PR #353, opened 2024-05-02
Explain +1

qwrtln:


.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -161,6 +161,8 @@ jobs:
           PAGES_PRINTABLE="${{ steps.check_printable_pages.outputs.num_of_printable_pages }}"
           echo "Number of pages in digital:   ${PAGES_DIGITAL}"
           echo "Number of pages in printable: ${PAGES_PRINTABLE}"
+          # If the number of pages in digital version is uneven, the printable one adds an index page.
```
DarthGandalf, line 164: I was sure index is added always. Did you mean notes? That means +1/+2 depending on evenness, but I see only +1 here

qwrtln: Well, it depends. If the number of pages is odd, just adding an index page makes the printable even. But if it's even, adding index page will make it uneven. In that case we'll need to add notes page to make it even. The goal is to have the number of pages in printable even at all times.

DarthGandalf: But the formula doesn't check for evenness at all

qwrtln: Yes, right now oddness is assumed, as it didn't change for a long time. That can be the next step.

DarthGandalf: Please mention it in the comment :)

qwrtln: Done.

DarthGandalf (approved):

README.md
```diff
@@ -10,9 +10,9 @@ Click in the table to download the most recent builds in the chosen language:
...
-| 🇪🇸 Español  |   ~80%   | [🇪🇸📜](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/es/main_es.pdf) | [🇪🇸🖨️](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/es/printable_es.pdf) |  [n/a](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/components_list_es/components_list_es.pdf)   |
-| 🇫🇷 Français |   ~40%   | [🇫🇷📜](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/fr/main_fr.pdf) | [🇫🇷🖨️](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/fr/printable_fr.pdf) |  [n/a](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/components_list_fr/components_list_fr.pdf)   |
+| 🇵🇱 Polski   |   870%   | [🇵🇱📜](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/pl/main_pl.pdf) | [🇵🇱🖨️](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/pl/printable_pl.pdf) |  [n/a](https://raw.githubusercontent.com/qwrtln/Homm3BG-build-artifacts/components_list_pl/components_list_pl.pdf)   |
```
DarthGandalf, line 13: 870%? really?

qwrtln: Neovim breaks with emoji 🤪 Fixed.

PR #359, opened 2024-05-02
[ES] printable bugs

luischos:
Finally, all translated correctly and well put together.


luischos:
Can't attach the PDFs, they are to heavy

luischos:
No, It's not, 52 pages, I have created it right now merged with main

luischos:
I think the problem is about the font style, or font size, from my Local and Git, is not the same, the PDFs are the same but haven't the same style, and this make an extra page.

luischos:
It's solved :D

qwrtln (approved):

PR #375, opened 2024-05-03
[FR] revision for difficulty, notes and trade

Gastonia02:
fr-37
fr-38
fr-39


qwrtln (changes requested):
Well that's weird. Why does it want to change so much? Which version of po4a do you have?

qwrtln:
When I run it on 0.71, it restores what you're doing here.

Gastonia02:
po4a 0.71, lots of changes in ru.po files

qwrtln (approved):

PR #388, opened 2024-05-05
Fix indentation for multiline index entries

qwrtln:
Look at "Karta Astrologowie Ogłaszają" and "Karta Ulepszonych Statystyk". Fixes the same problem for French too.
before|after


qwrtln:
> Picture or didn't happen

Patience, takes a while. Haven't scripted that yet.

DarthGandalf (approved):

evilmint (approved):

PR #410, opened 2024-05-07
Make setup image visible with enumeration

qwrtln:
The idea is to put the image in between the enumeration (before|after):

en-06
en-07
en-08

### DIFF
--- a/sections/setup.tex
+++ b/sections/setup.tex
@@ -45,7 +45,7 @@ This section will guide you through the process of setting up a Scenario from th
 \end{multicols*}
 
 \vspace*{\fill}
-\begin{figure}[!h]
+\begin{figure}[!ph]
   \centering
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Have you tried `\begin{figure}{p}` to let it automatically put the image to the separate page without forcing the exact split of the list?

qwrtln:
> Have you tried `\begin{figure}{p}` to let it automatically put the image to the separate page without forcing the exact split of the list?

I'm not sure what you mean by that? I can make it `[!ph]` to avoid a page break, but you still have to make it two separate lists. The list is in `multicols` environment, while the image is not.

DarthGandalf:
No, I meant literally `{p}` and keeping a single list.

I'll play around with this after work this evening

PR #417, opened 2024-05-07
Another compare script

DarthGandalf:


DarthGandalf:
Perhaps I need to replace `before` and `after` with `a` and `b` for A/B test, and it's easier to type than the long words.

Also perhaps I need to store all of this in a subdirectory of /tmp/

README.md
```diff
@@ -181,6 +181,30 @@ It will produce files: `en-38.png`, `en-39.png`, `en-40.png` and `en-41.png`.
...
+
+If you have only 1 page to show, just type the single number, without `-`. Or use a comma-separated list of numbers.
+
+The script will save the image of the specified pages to /tmp/. Now switch to the branch you're working on, do some changes, don't forget to rebuild the .pdf, and run:
+
+```bash
+tools/compare_pages_2.sh en 38-41 after
+```
+
+The script will use the previously written "before" files as the left side, and produce the single image with all the pages combined side by side, named /tmp/en-both-38-41.png
```
qwrtln, line 202: Can you make it so that it doesn't leave leftovers? I don't like the idea of having to remember to clean up after the script manually.

README.md
```diff
@@ -181,6 +181,30 @@ It will produce files: `en-38.png`, `en-39.png`, `en-40.png` and `en-41.png`.
 
 **This script requires `pdftoppm` and `imagemagick` utilities.**
 
+### 🎭 Comparing two pages side by side (v2)
+
+Another version, which works faster, but uses a different workflow:
+
+First, switch to the `main` branch, build the `main_en.pdf` and run:
```
qwrtln, line 188: Both scripts need this file. How about storing them online with the PDFs? Upon merging to main, pictures could be created and uploaded to GitHub.

README.md
DarthGandalf: These "leftovers" is the whole point of the optimization here, to avoid extract the same image again and again.

But if these leftovers are in somewhere like /tmp/homm3bgrecompare/, I don't think it's a big deal.

But feel free to reject this PR, I can just keep using it myself without sharing.

README.md
DarthGandalf: It's easy enough to extract the image from pdf, this is done with a single command. But downloading artifacts from github is slower/more complicated - needs github API tokens, etc.

README.md
DarthGandalf: OTOH, since it's more than twice faster than the other script, probably it would make sense to rewrite that one to use qpdf in similar way, maybe, even if extracting from both pdfs, and removing leftovers

README.md
qwrtln: No it's not. It would be the same as downloading PDFs from readme. On my machine, it takes ~15 s to build the PDF. Say you want to compare 3 images, that's 6 MB? I'll download it faster than in 15 s.

DarthGandalf: Ah, you mean not an artifact, but the separate repo. Then... maybe, but redownloading the same image again and again is still waste of bandwidth

qwrtln: Well, I can afford to download 6 MB from time to time 😆 Doesn't seem like a huge cost.

README.md
qwrtln: Will it produce the same image quailty?

DarthGandalf: Not sure. But I recall you were saying the purpose of the images was to ensure we don't overflow pages. The produced quality is enough for that

qwrtln: If I have a choice of a nice picture in 4s, and a grainy picture in 2s, I prefer the former :P

DarthGandalf: I passed the -density param to montage to make it the same quality as the other script

README.md
tomaas-zeman: I would also prefer downloading the files - full build on my machine takes 45s which is crazy long (twice as much if I also want printable).

But downloading the latest build for the given language and caching it for a while would make the workflow easier and I wouldn't have to keep track of the baseline builds. It could also have an option to download/compare regular and `printable` versions.

tomaas-zeman: This is my attempt on this topic https://github.com/Heegu-sama/Homm3BG/pull/567.

I use it for myself but if you're interested, I can make it a proper PR

tomaas-zeman: I guess this PR can be closed since the features were implemented in https://github.com/Heegu-sama/Homm3BG/pull/567?

PR #441, opened 2024-05-11
Fit for 56

qwrtln:
The printable document has 54 pages. It should have 56 (be divisible by 4) to make it possible to print it as a handy booklet (4 pages per A3 sheet, folded).
This change introduces an empty page with art after Combat section, and one-page Notes section after All Map Locations to achieve that. Benefits: three 2-page sections (AI Rules, Difficulty, Expansion Content) now become visible at once when opened as a booklet.

quote

ai

difficulty

expansion

notes

And as a bonus digital version with the quote:

en-02


metadata.tex
```diff
@@ -294,6 +294,8 @@
 
 \include{\sections/combat.tex}
 
+% Empty page placeholder
```
DarthGandalf, line 297:
```suggestion
\iftoggle{printable}{\include{....}}{}
```

DarthGandalf: I was going to cleanup Notes after release, but since you want to add yet another empty page, there's no reason to use this way for it from the start

qwrtln: Thanks for the refresher.

DarthGandalf (approved):

PR #443, opened 2024-05-11
Use more beautiful bullet point, from original rules

DarthGandalf:
en-both-11-13,17,18,23,25,28-29,31,34,37,39,41-42


qwrtln:
I like the idea, but may I suggest making it a little bigger and using an upscaled version of the dot?
image

qwrtln:
This teeny-tiny dot doesn't look good compared to the black one. If we want to use something similar to the original, I suggest we use a similar size.

bullet_point

qwrtln:
Also, please put it in `assets/layout` instead of `assets/images`.

DarthGandalf:
Ah, that's better. I made it so small because otherwise it was too pixelated

qwrtln (approved):

Heegu-sama (approved):

PR #450, opened 2024-05-13
Use ghostscript to compress workflow PDFs

qwrtln:
This reduces PDF size significantly.

### DIFF
--- a/.github/workflows/build-and-publish-rule-book.yaml
+++ b/.github/workflows/build-and-publish-rule-book.yaml
@@ -98,7 +98,7 @@ jobs:
 
       - name: Compress PDF
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -142,6 +155,13 @@ jobs:
             for _ in {1..2}; do ${{ matrix.luaengine && 'lualatex' || 'pdflatex' }} --shell-escape '\AtBeginDocument{\toggletrue{printable}} \input{main_${{ matrix.language }}.tex}'; done
           root_file: main_${{ matrix.language }}.tex
 
+      - name: Optimize PDF
```
DarthGandalf, line 158: I didn't know you can have 2 steps with the same name

PR #469, opened 2024-05-13
Improve printable index

DarthGandalf:


qwrtln (changes requested):
Doesn't work in the workflow.

qwrtln:
This is the index page from the PDF built in the workflow:

en-55

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -133,13 +134,9 @@ jobs:
         uses: xu-cheng/latex-action@v3
         with:
           extra_system_packages: "inkscape"
+          pre_compile: export HOMM3_PRINTABLE=1 HOMM3_LANG=${{ matrix.language }}
```
qwrtln, line 137:
```suggestion
        uses: xu-cheng/latex-action@v3
        env:
          HOMM3_PRINTABLE: 1
          HOMM3_LANG: ${{ matrix.language }}
        with:
          extra_system_packages: "inkscape"
```

qwrtln: Not sure if it works on actions, though

qwrtln: If that doesn't work, I guess you can just do 
```
echo "HOMM3_PRINTABLE=1" >> $GITHUB_ENV
```
in a previous step.

DarthGandalf:
Fixed the empty index.

Caveat: I simply took the first character, and compared their codepoints, ignoring existence of collation, so for some languages the result may be wrong. This will be easier to fix when such cases appear

qwrtln:
Ukrainian index doesn't look good. Does it work for Cyrillic alphabet?

image

qwrtln (approved):
Perfect

PR #482, opened 2024-05-15
[UA] AI Rules

SerhiiSUr:
new example


qwrtln (changes requested):
Please edit the existing file. Adding new one has no effect.

PR #485, opened 2024-05-15
Format quote properly

qwrtln:
English:
en-37

French:
fr-37

Polish:
pl-37

Ukrainian:
ua-37


DarthGandalf:
I'd probably plumb it through .po instead - it's not that different from a "section" except that it's included in 2 places and doesn't have a section header

qwrtln:
> I'd probably plumb it through .po instead - it's not that different from a "section" except that it's included in 2 places and doesn't have a section header

Good idea, wanted to implement it, but sadly po4a doesn't support it: https://github.com/mquinson/po4a/issues/405

Heegu-sama (approved):

SerhiiSUr (approved):

PR #489, opened 2024-05-16
Ranged clarification

Heegu-sama:
Added the word "enemy" to the ranged unit description so as to not imply that they can attack friendlies directly.


qwrtln:
po4a commit

qwrtln (approved):

PR #491, opened 2024-05-16
Update ua.po

SerhiiSUr:
Ranged Units update


qwrtln (changes requested):
Remove the fuzzy entry

qwrtln (approved):

PR #501, opened 2024-05-20
Readme for 1.2

qwrtln:
Standard before|after picture below. I'm open to suggestions regarding emojis.
image


evilmint:
imo there's a mess with the emojis, maybe instead of the col+row combination in the cells, just put a <a href="...">Download</a> link ?

qwrtln:
> imo there's a mess with the emojis, maybe instead of the col+row combination in the cells, just put a [Download](...) link ?

Fixed

JStanula:
I would put something like "stable_PL" "beta_EN"

qwrtln:
> I would put something like "stable_PL" "beta_EN"

Where exactly? Can you give me 4 names for EN and PL each?

JStanula:
I mean instead of "download"

DarthGandalf (approved):

PR #523, opened 2024-05-21
[FR] Adjusting paralysis token

Legrecofr:
Better layout.

fr-06

fr-06


qwrtln:
I made it so that your change is visible in the digital build, but left the old way for printable, as it looked better there.

qwrtln (approved):

PR #536, opened 2024-05-23
Make all links pagelinks

DarthGandalf:


qwrtln (commented):
Maybe instead of changing hyper->page everywhere, just redefine hyper? And use something like "coverlink" for the three exceptions.

DarthGandalf:
If hyper is redefined, how to actually do the hyperlink? What would be the definition of it?

qwrtln:
> If hyper is redefined, how to actually do the hyperlink? What would be the definition of it?

Ah, true, we can't just overload it.

qwrtln:
Please fix the link which refers to the field difficulty table (p. 56). Can't remember where it was exactly, Units?

DarthGandalf:
> Please fix the link which refers to the field difficulty table (p. 56). Can't remember where it was exactly, Units?

I can't find any more. I only fixed the one in the difficulty section.

The Combat section links to Difficulty section, but not the difficulty table

sections/heroes.tex
```diff
@@ -12,21 +12,21 @@ \subsection*{Main Hero}
 Each Faction's Main Hero has 3 Movement Points.
 Only the Main Hero can use Your Deck.\par
 Each Main Hero starts the game at Level 1 and can advance up to Level 7 by gaining Experience.
-Experience is gained from \hyperlink{Combatexperience}{winning Combat}, Visiting certain \hyperlink{All Map Locations}{Locations} and the \hyperlink{Treasure Die}{Treasure Die} \includesvg[height=10px]{\svgs/treasure.svg}.
+Experience is gained from \pagelink{Combatexperience}{winning Combat}, Visiting certain \pagelink{All Map Locations}{Locations} and the \pagelink{Treasure Die}{Treasure Die} \includesvg[height=10px]{\svgs/treasure.svg}.
```
qwrtln, line 15:
```suggestion
Experience is gained from \pagelink{Combatexperience}{winning Combat}, Visiting certain \pagelink{All Map Locations}{Locations} and the \pagelink{Treasure Die}{Treasure Die \includesvg[height=10px]{\svgs/treasure.svg}}.
```
I think it's better to make the glyph a part of the link, so that it's not rendered behind the page number.

image

qwrtln (approved):

Heegu-sama (approved):

PR #540, opened 2024-05-26
Fix some typos

qwrtln:


evilmint (approved):

translations/expansion_content.tex/pl.po
```diff
@@ -190,7 +190,7 @@ msgid ""
 msgstr ""
 "\\subsection*{\\pagetarget{War Machines}{Machiny Wojenne}}\n"
 "Dodane w rozszerzeniu Bastion.\n"
-"Machiny Wojenne\\index{Machiny Wojenne} są Kartami z efektem Długotrwałym, można je zakupić w \\pagelink{Trading Post}{Punkcie Handlowym} lub \\pagelink{War Machine Factory}{Fabryce Machin Wojennych}.\n"
+"Machiny Wojenne\\index{Machiny Wojenne} są Kartami z efektem Długotrwałym, można je zakupić w \\hyperlink{Trading Post}{Punkcie Handlowym} lub \\pagelink{War Machine Factory}{Fabryce Machin Wojennych}.\n"
```
DarthGandalf, line 193: why?

DarthGandalf (approved):

translations/expansion_content.tex/pl.po
qwrtln: Because it's the same page repeated in the same line. No point in shortening the text for it, because the hyperlinks refers to the same place anyway (two pictures on the same horizontal position).

qwrtln: Just like the paralysis/defense token, same case.

DarthGandalf: Then English and other translations probably should be updated as well

PR #541, opened 2024-05-26
[FR] fix linkable

Gastonia02:


Gastonia02:
po4a check

Gastonia02:
po4a commit

PR #544, opened 2024-05-29
Delete table drop shadow from back cover tables

qwrtln:
This shadow looks badly when rendered for printing. This PR adds an option to make it fully transparent, effectively getting rid of it.

See the diff (before|after):

en-53


DarthGandalf:
Is the shadow good for other tables on other pages?

qwrtln:
> Is the shadow good for other tables on other pages?

It is. We're using Adobe's tech to add margins and crop markings for the print shop. I guess we could add them in LaTeX too and make another version for professional printing. Anyhow, for whatever reason, Adobe struggles to render those shadows:

image

Other shadows look good, though. I think it's best to remove those faulty ones, as they do not improve aesthetics of the page anyway.

Heegu-sama (approved):

evilmint (approved):

qwrtln:
Nope, other shadows also have this artifact. It's less visible though on the pale background. Adobe clearly doesn't like LateX shadows. One workaround is to render pages as TIFF images. I still think those shadows on the last page is a good riddance. What do you think @DarthGandalf @evilmint @Heegu-sama?

qwrtln:
> I don't mind removing them everywhere if that helps

The artifacts are slightly visible. How about we delete the shadows from the printable build only?

image

qwrtln:
While keeping them deleted from the back cover in both variants.

DarthGandalf (approved):

PR #552, opened 2024-06-02
Adjust two-side layout for printable

qwrtln:
The purpose of this change is to avoid this effect (look how close the text is to the inner margins), no change to the digital build:

image

This change aims to shift margins for the printable build, so that inner sides of the pages leave some space for glueing/binding. All pages apart from 1 and 56 were moved 5 mm "outwards". This image visualizes it well:

image


Heegu-sama (approved):

sections/components.tex
```diff
@@ -1,6 +1,8 @@
 % !TeX spellcheck = en_US
 \addsection{Components}{\skills/luck.png}
 
+\iftoggle{printable}{\vspace{-\baselineskip}}{}
```
DarthGandalf, line 4: shouldn't this be part of \addsection?

DarthGandalf: That can also simplify all the translations

DarthGandalf (approved):

sections/components.tex
qwrtln: I wish it could be, but it's inconsistent, and this problem doesn't appear everywhere. Changing to `twoside` messes with the margins and seems to insert a newline randomly.

qwrtln: And it doesn't impact translations at all, so no worries here.

JStanula:
How about printing on A3 paper and binding it by office staple in the middle like it is done in the official manual?? 
I think this idea with extra margin is good but IMHO it should be an optional rather than fixed option.

qwrtln:
> How about printing on A3 paper and binding it by office staple in the middle like it is done in the official manual?? I think this idea with extra margin is good but IMHO it should be an optional rather than fixed option.

It's experimental for now. You have the stable build if you need the old layout. In any case, a PR is welcome.

PR #553, opened 2024-06-03
Build printable version without background

tomaas-zeman:
Addresses https://github.com/Heegu-sama/Homm3BG/issues/546

I didn't touch the release flow though. There seemed to be multiple steps related to generating the current printable version so I didn't want to unecessarily duplicate that part

PR #555, opened 2024-06-05
Refactor version handling during build and release

tomaas-zeman:
Based on the previous discussion https://github.com/Heegu-sama/Homm3BG/pull/553#discussion_r1626067929

- version is stored in `.version` file and not parsed from the document
- version number is then injected during build
- ~~versions are also replaced in `README` links during release~~


tools/release.sh
```diff
@@ -42,4 +42,10 @@ echo "Optimizing economy printable build..."
 tools/optimize.sh ${LANGUAGE} cmyk &> /dev/null
 mv main_${LANGUAGE}_optimized.pdf release-${VERSION}/Heroes3_${languages[$LANGUAGE]}_Rules_Rewrite_${FILE_VERSION}_Economy_Printable.pdf
 
+echo "Updating README links..."
+sed -E -i.tmp "s;Version [.0-9]+;Version ${VERSION};" README.md
+sed -E -i.tmp "s;releases/download/v[^/]+;releases/download/v${VERSION};" README.md
+sed -E -i.tmp "s;Rules_Rewrite_[0-9]+;Rules_Rewrite_${FILE_VERSION};" README.md
```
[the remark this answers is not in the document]
DarthGandalf: Make it `1.2`?

I imagine we can potentially go above version 9, which would be 10, and eventually have the same name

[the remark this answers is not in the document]
qwrtln: Initially, we avoided dots in the file names. Do all systems support it?

metadata.tex
```diff
@@ -53,6 +54,13 @@
 \usepackage[skip=6pt, indent=0pt]{parskip}
 \setstretch{1}
 
+% Get version from env
+% \getenv{variable_name} just prints the value
+% \getenv[\macro]{variable_name} stores the value in \macro for reusability
+\newcommand{\getenv}[2][]{%
+  \CatchFileEdef{\value}{"|kpsewhich --var-value #2"}{\endlinechar=-1}%
+  \if\relax\detokenize{#1}\relax\value\else\let#1\value\fi}
```
DarthGandalf, line 62: Why do you need this over simple `\input{.version}`?

qwrtln:
This will not work without refactoring the workflow.

metadata.tex
```diff
@@ -53,6 +54,13 @@
 \usepackage[skip=6pt, indent=0pt]{parskip}
 \setstretch{1}
 
+% Get version from env
+% \getenv{variable_name} just prints the value
+% \getenv[\macro]{variable_name} stores the value in \macro for reusability
+\newcommand{\getenv}[2][]{%
+  \CatchFileEdef{\value}{"|kpsewhich --var-value #2"}{\endlinechar=-1}%
```
DarthGandalf, line 61: Sorry, why do you need to run external command (part of kpathsea) for this?

qwrtln:
> Based on the previous discussion [#553 (comment)](https://github.com/Heegu-sama/Homm3BG/pull/553#discussion_r1626067929)
> 
>     * versions are also replaced in `README` links during release


I'm not sure if it's necessary. Readme evolves every time there's a release. We're not at the point where we need automation here. If anything, it should be a separate script, because I don't want to edit README when I build the release files. And you run it once for every language. And no, I don't want it to build every language on a single run, as a simple bash for loop will do. I want the capability to build a single language.

main_en.tex
```diff
@@ -8,9 +8,7 @@
 \def\sections{sections}
 \def\lang_header_adjustment{0px}
 
-\newcommand{\version}{
-  Version 1.2
-}
+\newcommand{\versionlabel}{Version}
```
qwrtln, line 11:
```suggestion
\newcommand{\version}{
  Version  \input{.version}
}
```
Something simple like this will not do?

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -25,51 +25,31 @@ jobs:
...
-            version: Version
-            warning: "Dies ist eine unveröffentlichte Version auf GitHub"
             sections: "sections/translated/de"
             luaengine: true
 
     steps:
       - name: Set up repository
         uses: actions/checkout@v4
 
-      - name: Replace rule book version with a warning
```
qwrtln, line 67: Why? We want this.

qwrtln: Ah, ok, you put it in the latex code.

metadata.tex
tomaas-zeman: That's good. I didn't know about that

The `\getenv` macro might be needed for this though, unless input also reads env variables

- https://github.com/Heegu-sama/Homm3BG/actions/runs/9379614635/job/25824911861
- https://github.com/Heegu-sama/Homm3BG/pull/555/commits/122d86e576ea5e486cf35a185e47cdd5accbc71c

.github/workflows/build-and-publish-rule-book.yaml
DarthGandalf: Thanks, I was going to move that warning over as well, to have all the strings (which are not in .po) translated in the same place

tomaas-zeman:
> > Based on the previous discussion [#553 (comment)](https://github.com/Heegu-sama/Homm3BG/pull/553#discussion_r1626067929)
> > ```
> > * versions are also replaced in `README` links during release
> > ```
> 
> I'm not sure if it's necessary. Readme evolves every time there's a release. We're not at the point where we need automation here. If anything, it should be a separate script, because I don't want to edit README when I build the release files. And you run it once for every language. And no, I don't want it to build every language on a single run, as a simple bash for loop will do. I want the capability to build a single language.



> > Based on the previous discussion [#553 (comment)](https://github.com/Heegu-sama/Homm3BG/pull/553#discussion_r1626067929)
> > ```
> > * versions are also replaced in `README` links during release
> > ```
> 
> I'm not sure if it's necessary. Readme evolves every time there's a release. We're not at the point where we need automation here. If anything, it should be a separate script, because I don't want to edit README when I build the release files. And you run it once for every language. And no, I don't want it to build every language on a single run, as a simple bash for loop will do. I want the capability to build a single language.

OK. I'll revert that change

metadata.tex
DarthGandalf: Currently environment is read in perl in latexmkrc

metadata.tex
tomaas-zeman: Sorry, I'm not latex pro and I didn't find a better way to read env variables

tools/release.sh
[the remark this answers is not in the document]
tomaas-zeman: Code removed

metadata.tex
DarthGandalf: Don't worry, me neither :D

metadata.tex
tomaas-zeman: Can I define any variable in the latexmkrc or do I have to squueze everything into the `pre_tex_code`?

.version
```diff
@@ -0,0 +1 @@
+1.2
```
qwrtln, line 1: I guess this file needs to be without the final newline in order not to confuse `cat` or whatever else?

main_en.tex
tomaas-zeman: done

.version
tomaas-zeman: How can I exclude it from the check?

metadata.tex
tomaas-zeman: Maybe there's an easier way (shellesc package, latexmkrc) but I was not able to make anything else work.

But I'm open to ideas :)

.version
qwrtln: Did you check this? I just tried it locally. Works even with the final newline.

metadata.tex
tomaas-zeman: I changed the `kpsewhich` command to simple `echo` to get the env variable. That should be more acceptable :)

metadata.tex
tomaas-zeman: Let's continue here https://github.com/Heegu-sama/Homm3BG/pull/555/files#r1627353364

metadata.tex
qwrtln: A controversial question: will it work on Windows?

.version
tomaas-zeman: Hmm, it does. That's quite odd. Maybe it would make sense to me in case of `\input` macro but certainly not for `cat`

tomaas-zeman: TIL: "Command substitutions strip all trailing newlines from the output of the command inside them"

https://mywiki.wooledge.org/CommandSubstitution

metadata.tex
```diff
@@ -278,7 +289,13 @@
 \maketitle
 
 \begin{center}
-  \version{}
+  \getenv[\githubsha]{GITHUB_SHA}
```
qwrtln, line 292: Gosh, I had no idea LaTeX had such capabilities. I'm a little uneasy with adding an extra package just for this, but on the other hand I like the parametrization, which simplifies the workflow.

.version
qwrtln: TIL too :D

metadata.tex
tomaas-zeman: Does it have to? I mean all scripts are written in bash :)

qwrtln: The owner doesn't work actively on this any more, but up until now you could just open the project in TeXStudio and build over there. I'm afraid that with this change it will become impossible.

qwrtln: OMG, it breaks at getenv, not here haha.

tomaas-zeman: Anyway, I'll follow your recommendations anyway on this topic but supporting Windows is just another burden to bear. So unless there's a good reason to do so (like at least few active devs), I would discourage that

tomaas-zeman: So if I go back to the previous command, I still don't have a way to test that it works properly. I could assume it'll work but still

metadata.tex
qwrtln: How about putting this under another iftoggle, which kicks in only on GitHub? It should allow us to bypass the TeXstudio error?

tomaas-zeman: That could work. How to detect the github env?

qwrtln: Just define another `iftoggle`, and pass it over in the `pre_compile:` arg of the LaTeX GitHub action.

qwrtln: Something like `githubbild`?

qwrtln: Exactly the same way that `printable` is passed, but make use it on GitHub only.

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -10,9 +10,9 @@ on:
 env:
   PO4A_VERSION: "0.71"
   PO4A_GH_URL: https://github.com/mquinson/po4a/releases/download
-  REPO_URL: https://github.com/Heegu-sama/Homm3BG
   GHOSTSCRIPT_URL: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs10030/gs_10.03.0_amd64_snap.tgz
   ARTIFACT_RETENTION_DAYS: 1
+  HOMM3_GITHUB_BUILD: 1
```
qwrtln, line 15: Didn't work. Please try putting it in the `pre_complie:` arg in every step which builds the PDF file.

tomaas-zeman: Yeah, I noticed. I kind of expected this would fail but still wanted to give it a try and avoid unnecessary duplication

qwrtln: Confirmed 👍🏻

qwrtln:
@csagataj2 You were the one? Please share some good news with us and tell us if it builds.

tomaas-zeman:
@qwrtln It seems that job config change also changed the structure of checks and the deprecated one are not running now, yet they still seem to be required

Screenshot 2024-06-05 at 15 36 09

qwrtln:
> @qwrtln It seems that job config change also changed the structure of checks and the deprecated ones are not running now, yet they still seem to be required
> Screenshot 2024-06-05 at 15 36 09

Required checks are picked manually in the repo settings, to which only the owner has access. Since the jobs evolved (matrix has different params now), old ones need to be deleted and the new ones selected. We shall do it once we have confirmation that everything's working as intended.

csagataj2:
> @csagataj2 You were the one? Please share some good news with us and tell us if it builds.

It works as intended

qwrtln:
> > @csagataj2 You were the one? Please share some good news with us and tell us if it builds.
> 
> It works as intended

I'm glad we're leaving this _window_ of opportunity open for everyone

DarthGandalf (approved):

qwrtln (approved):

qwrtln:
@Heegu-sama We need your help here. Those yellow required checks which won't run now need to go away. They should be replaced by those which run successfully, but are not labelled as required.

Heegu-sama (approved):

PR #557, opened 2024-06-05
Initial Czech language commit

tomaas-zeman:


qwrtln:
Please give me the text you'd like in place of "RULE BOOK REWRITE" on the cover in Czech. I'll create a suitable image for you.

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -59,6 +59,11 @@ jobs:
             warning: "Dies ist eine unveröffentlichte Version auf GitHub"
             sections: "sections/translated/de"
             luaengine: true
+          - language: cs
+            version: Verze
+            warning: "Toto je nepublikovaná verze vyrobená z commitu na GitHubu"
+            sections: "sections/translated/cs"
+            luaengine: true
```
qwrtln, line 66:
```suggestion
            luaengine: false
```
We're using LuaLaTeX only for Cyrillic languages. The German version with it is an error.

main_cs.tex
```diff
@@ -0,0 +1,33 @@
+% !TeX spellcheck = cs_CZ
+\documentclass[12pt]{article}
+
+% Language setting
+\usepackage[T1]{fontenc}
+\usepackage[english]{babel}
```
qwrtln, line 6:
```suggestion
\usepackage[czech]{babel}
```

main_cs.tex
```diff
@@ -0,0 +1,33 @@
...
+
+\def\sections{sections}
```
qwrtln, line 8:
```suggestion
\def\sections{sections/translated/cz}
```

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -59,6 +59,11 @@ jobs:
             warning: "Dies ist eine unveröffentlichte Version auf GitHub"
             sections: "sections/translated/de"
             luaengine: true
+          - language: cs
+            version: Verze
+            warning: "Toto je nepublikovaná verze vyrobená z commitu na GitHubu"
+            sections: "sections/translated/cs"
```
qwrtln, line 65:
```suggestion
```
Actually, after recent changes, this line is no longer required. Can you fix it in #555?

tomaas-zeman:
> Please give me the text you'd like in place of "RULE BOOK REWRITE" on the cover in Czech. I'll create a suitable image for you.

I think you can put "Revidovaná pravidla" there. Sadly, there isn't consistency between translations

.github/workflows/build-and-publish-rule-book.yaml
tomaas-zeman: Do you have a PR where this was introduced? Just for a reference

qwrtln:
This goes to `translations/assets/cs`
cover_subtitle

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Not sure why you would need that? Just delete those lines.

tomaas-zeman: Just to understand what I'm doing :)

tomaas-zeman:
> Changing the workflows requires updating them in the settings every time the change is made. Please, don't wait with the change, and do it once in a single place. Let's respect each other's time. Get rid of the leftovers in #555.

Got it, it's done

.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -59,6 +59,10 @@ jobs:
             warning: "Dies ist eine unveröffentlichte Version auf GitHub"
             sections: "sections/translated/de"
             luaengine: true
+          - language: cs
+            version: Verze
```
tomaas-zeman, line 63: This needs to wait for #555 to be merged and moved to main_cs.tex

DarthGandalf:
> We're using LuaLaTeX only for Cyrillic languages. The German version with it is an error.

For Cyrillic that's a necessity, but for Latin it's more a preference. The default LaTeX font was producing very bad result for headers in Russian, so we switched to TTF font which didn't have such issues, but that requires Lua. Really there's nothing preventing using Lua for other languages as well.

DarthGandalf (approved):

sections/title_page.tex
```diff
@@ -6,7 +6,7 @@
       \includegraphics[height=\paperheight, keepaspectratio]{\layout/cover.jpg}
     };
   }
-  \node(title)[minimum width = \paperwidth, anchor=center, yshift=-10em] at (current page.north) {
+  \node(title)[minimum width = \paperwidth, anchor=center, yshift=\dimexpr-10em\relax] at (current page.north) {
```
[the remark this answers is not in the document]
tomaas-zeman: `\setlength` also worked. I'm not sure if there is another way to wrap a number and pass it there so that it works with that babel

.github/workflows/build-and-publish-rule-book.yaml
tomaas-zeman: ✅

qwrtln (approved):

qwrtln (changes requested):
Oops, there's another job in the workflow which checks if the number of pages is correct. Please add it for your language.

tomaas-zeman:
> Oops, there's another job in the workflow which checks if the number of pages is correct. Please add it for your language.

Sorry, I missed that one. Fixed

qwrtln:
> > Oops, there's another job in the workflow which checks if the number of pages is correct. Please add it for your language.
> 
> Sorry, I missed that one. Fixed



> > Oops, there's another job in the workflow which checks if the number of pages is correct. Please add it for your language.
> 
> Sorry, I missed that one. Fixed

Perfect. @Heegu-sama, we need your help again with adding Czech to the mandatory checks.

qwrtln (approved):

Heegu-sama (approved):

qwrtln (changes requested):
Sorry, until the checks are mandatory, I'm blocking the PR. In every language there should be 53 pages.

tomaas-zeman:
> Sorry, until the checks are mandatory, I'm blocking the PR. In every language there should be 53 pages.

This is because of czech babel again. Some of the layouts got broken because words are differently wrapped. I can hotfix this but I'm slightly afraid this could pop up again in the future.

Screenshot 2024-06-06 at 21 56 19

In this case the issue is valid because it really looks ugly. But what if, due to language nature, we need more pages in Czech to explain the rules?

qwrtln:
> > Sorry, until the checks are mandatory, I'm blocking the PR. In every language there should be 53 pages.
> 
> what if, due to language nature, we need more pages in Czech to explain the rules?

Literally every other translator asked exactly the same question. And guess what: in every language, the document has 53 pages. When in doubt, refer to the original, which also maintains the same number of pages and the same layout for every language.

qwrtln (approved):
Checks are mandatory now. Apart from the layout, everything looks good.

tomaas-zeman:
> > > Sorry, until the checks are mandatory, I'm blocking the PR. In every language there should be 53 pages.
> > 
> > 
> > what if, due to language nature, we need more pages in Czech to explain the rules?
> 
> Literally every other translator asked exactly the same question. And guess what: in every language, the document has 53 pages. When in doubt, refer to the original, which also maintains the same number of pages and the same layout for every language.

I'm saying that fixing three pages now is easy. I'm not sure how many changes had to be made in other languages but I just don't want to end up in situation where a new chapter is added and `czech` will not pass this test. This could be a pain for all future contributors.

I have no idea how often that happens so I'm obviously very cautious about it

qwrtln:
> I'm saying that fixing three pages now is easy. I'm not sure how many changes had to be made in other languages but I just don't want to end up in situation where a new chapter is added and `czech` will not pass this test. This could be a pain for all future contributors.

Believe me, translating crews in every language worked really hard on this. The problem is not unique to Czech language, every translation carries that risk.

qwrtln:
As a workaround, I guess you could use both Czech and English for babel now?

DarthGandalf:
We don't add new chapters very often. But I agree, the layout is pretty fragile, but that's why there are automated checks here for every PR

tomaas-zeman:
> As a workaround, I guess you could use both Czech and English with babel for now?

Yeah, that's exactly what I'm trying now but it doesn't seem to help much

tomaas-zeman:
> Believe me, translating crews in every language worked really hard on this. 

I don't doubt that :) My only concern was to not create annoying situation for others in future contribution if that was not clear

DarthGandalf:
As a workaround you can add `\vspace{-10em}` to places where layout broke, in the "translated" .po files, even with English text for now

qwrtln:
> As a workaround you can add `\vspace{-10em}` to places where layout broke, in the "translated" .po files, even with English text for now

I hate it, but it's probably the easiest solution for now xD

tomaas-zeman:
Concrete example. Even with multilanguage babel, the page doesn't get better :/ 

Screenshot 2024-06-06 at 22 18 13

qwrtln:
> I don't doubt that :) My only concern was to not create annoying situation for others in future contribution if that was not clear

Yeah OK, I should have communicated it more effectively or more politely, sorry. But the sad reality is that if a translation breaks the layout, then the translator hasn't done a very good job and needs to improve their work. Most translators just want to do it once and be done with it. It will be your job, the CEO of Czech translation, to keep them aware.

...which made me realize that the CODEOWNERS file was not updated with the Czech language. Please do it now.

tomaas-zeman:
Is there a way to force english babel to untranslated pages? Maybe the multilang doesn't work that well by default

qwrtln:
> Is there a way to force english babel to untranslated pages? Maybe the multilang doesn't work that well by default

Dump multilang, use `vspace`

qwrtln:
> Concrete example. Even with multilanguage babel, the page doesn't get better :/
> Screenshot 2024-06-06 at 22 18 13

Oh, it's clear and simple. The `subsection` spacing in Czech is bigger than in English. That's the problem. The answer for now is definitely `vspace`s.

DarthGandalf:
example of this workaround: https://github.com/Heegu-sama/Homm3BG/blob/28072d891b8df4a8df85f1912a59d62219a0110a/translations/components.tex/es.po#L34

tomaas-zeman:
> Yeah OK, I should have communicated it more effectively or more politely, sorry. 

No worries. Same apology from my side if I came across impolite.

> But the sad reality is that if a translation breaks the layout, then the translator hasn't done a very good job and needs to improve their work.

I see. If that's the world translators live in, I guess I'll go with the flow and try to do it so it doesn't break :D 
 
> ...which made me realize that the CODEOWNERS file was not updated with the Czech language. Please do it now.

Sure

tomaas-zeman:
> Oh, it's clear and simple. The `subsection` spacing in Czech is bigger than in English. That's the problem. The answer for now is definitely `vspace`s.

You're right! I was so focused on the different word wrap that I didn't noticed that

DarthGandalf:
> Unknown owner on line 50: make sure @tomaas-zeman exists and has write access to the repository 

hmm

tomaas-zeman:
> > Unknown owner on line 50: make sure @tomaas-zeman exists and has write access to the repository
> 
> hmm



> > Unknown owner on line 50: make sure @tomaas-zeman exists and has write access to the repository
> 
> hmm

But it doesn't seem to be mandatory for this PR 🤷

PR #559, opened 2024-06-05
Rephrase wrongly understood rule about playing attack/defense

DarthGandalf:
en-both-32

### DIFF
--- a/sections/deckbuilding.tex
+++ b/sections/deckbuilding.tex
@@ -118,9 +118,9 @@ Artifacts can be \pagelink{Trading}{traded} in Alliance and Cooperative Scenario
 Spell Cards\index{Spell Card} have three possible primary effects.
 Using the topmost, basic version of the Spell has no additional costs.
 To access the other effects, you may \textbf{Empower} a Spell\index{Empowering Spells} by paying the indicated cost (3) to get a more powerful outcome (4).
-You may pay this cost by playing other cards for their Empower \includesvg[height=10px]{\svgs/empower.svg} effect before casting the Spell.
-All Spell Cards also have an alternative bottom (5) \includesvg[height=10px]{\svgs/empower.svg} effect.\par
-
+You may pay this cost by playing other cards for their Empower \includesvg[height=10px]{\svgs/empower.svg} effect (e.g. the Spell Power Statistic) before casting the Spell.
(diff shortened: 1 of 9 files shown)


sections/combat.tex
```diff
@@ -82,7 +82,7 @@ \subsection*{\pagetarget{CombatCards}{Using Cards During Combat}}
 Ongoing \includesvg[height=10px]{\svgs/ongoing.svg} and \includesvg[height=10px]{\svgs/activation.svg} Activate effects can be used only \textbf{when Activating one of your Units and before it attacks}.
 Ongoing effects\index{Ongoing Effects} last until end of Combat or if the effect on the card is used up.\par
 Instant \includesvg[height=10px]{\svgs/instant.svg} Cards may be played \textbf{at any time} except between rolling the Combat Die and resolving damage unless otherwise stated.
-Instant Cards which increase a Unit's \includesvg[height=10px]{\svgs/attack.svg} or \includesvg[height=10px]{\svgs/defense.svg}, such as the Attack and Defense Statistic cards, \textbf{must} be played after an attack has been announced and persist only for the duration of that attack.
+Effects of Instant Cards, such as the Attack and Defense Statistic cards, which increase a Unit's \includesvg[height=10px]{\svgs/attack.svg} or \includesvg[height=10px]{\svgs/defense.svg}, expire whenever an attack happens or the Activation ends, whichever comes first.
```
tomaas-zeman, line 85: Does "attack happens" include even the retaliation phase?

Imagine this edge case.
1. I decide to attack and use +1 defense card
2. I roll the attack die and hit the enemy creature
3. Enemy creature now retaliates
4. Do I still have +1 defense on my creature for the retaliation attack or not?

I know it's a bit weird to use defense card before my attack but that would make me better understand the time frame of card effects

qwrtln: My guess would be no, you don't have +1 defense.

tomaas-zeman: I would think so too.

I would maybe use word "resolves" instead of "happens". Otherwise the change sounds good to me

DarthGandalf: Good question.

Выделение_577

I need to think how to phrase this.

DarthGandalf: Done. PTAL

qwrtln: LGTM. Nothing left to guesses now.

qwrtln (approved):

Heegu-sama (approved):

PR #564, opened 2024-06-07
[CS] game modes

tomaas-zeman:
en-05


qwrtln (approved):
Campaign looks unnecessarily lengthy, but other than that all good.

tomaas-zeman:
> Campaign looks unnecessarily lengthy, but other than that all good.

A tiny bit yes. The English version has two phrases `single player mode` and `Solo Mode` which sounded pretty much the same to me so I translated them both as `hra pro jednoho hrace` which is on the longer side. But if feels more natural to read

PR #565, opened 2024-06-07
[CS] components

tomaas-zeman:
cs-06
cs-07


DarthGandalf (approved):

qwrtln:
I suggest translating this section relying only on the printable version.

tomaas-zeman:
@qwrtln The text here talks about Archon content guide. Do we want to provide link to it? 
```
https://archon-studio.com/downloads/heroes-iii
```

qwrtln (approved):

PR #572, opened 2024-06-10
[CS] deckbuilding

tomaas-zeman:
cs-18
cs-19
cs-20
cs-21


tomaas-zeman:
Not sure if it's bad rendering on my side or a general issue but the note on the third page doesn't look good. I'll check once built in GitHub
Screenshot 2024-06-10 at 19 01 52

DarthGandalf (approved):

PR #575, opened 2024-06-10
[CS] town

tomaas-zeman:
cs-23


translations/town.tex/cs.po
```diff
@@ -78,6 +78,27 @@ msgid ""
 "Some Building Tiles are double-sided, and may later be upgraded and flipped to represent two different buildings at the same time. Such upgrades must be \\textbf{Built in order}.\\par\n"
 "\n"
 msgstr ""
+"\\pagetarget{Town}{Každá}\\index{Town} Frakce má své vlastní Město, umístěné uprostřed jejich Startovního Dílku.\n"
```
DarthGandalf, line 81: \index should be translated too - that's what goes to Index chapter at the end of printable pdf

DarthGandalf (approved):

translations/town.tex/cs.po
tomaas-zeman: Oh. I might have missed that in all my translations. Let me check

PR #579, opened 2024-06-11
Test unforgiving dread knight

qwrtln:
The goal is to disallow dread knight shrinking, which allowed some overflows to pass page checks.

before|after
en-35


DarthGandalf:
Image please

qwrtln:
> Image please

It's WIP, sorry

qwrtln:
Verbose translations will no longer be able to bypass the checks.

sections/combat.tex
```diff
@@ -204,12 +204,10 @@ \subsection*{Campaign and Cooperative Combat}
...
 
 \vspace*{\fill}
-\begin{scaledfigure}[blanker]
-  \centering
-  \includegraphics[width=\linewidth, height=\myspace, keepaspectratio]{\art/dread_knight.jpg}
-\end{scaledfigure}
 
-\clearpage
+\begin{center}
+  \includegraphics[width=0.97\linewidth]{\art/dread_knight.jpg}
```
tomaas-zeman, line 209: Not a big deal but to my surprise, I was able to squeeze Czech translation to a point where the Dread knight can be at full width, which looks slightly better. Maybe we could do the same with other languages

Screenshot 2024-06-11 at 13 45 35

qwrtln:
Agreed, but currently impossible without changing original text.

11 Jun 2024 13:46:48 Tomas Zeman ***@***.***>:

tomaas-zeman (approved):

qwrtln:
I have no idea why, but German yields different results locally and in the workflow. Apart from the word "ends" carried over to the next column, everything else seems identical.

@DarthGandalf @tomaas-zeman 

image

tomaas-zeman:
> I have no idea why, but German yields different results locally and in the workflow. Apart from the word "ends" carried over to the next column, everything else seems identical.

Isnt it because workflow has lua enabled and locally its disabled?

qwrtln:
> > I have no idea why, but German yields different results locally and in the workflow. Apart from the word "ends" carried over to the next column, everything else seems identical.
> 
> Isnt it because workflow has lua enabled and locally its disabled?



> > I have no idea why, but German yields different results locally and in the workflow. Apart from the word "ends" carried over to the next column, everything else seems identical.
> 
> Isnt it because workflow has lua enabled and locally its disabled?

Could be. I changed the workflow locally but haven't pushed yet. You can see that lua performs hyphenation in different places.

qwrtln:
po4a check

qwrtln:
@DarthGandalf What do you think? It should prevent any future abuse.

DarthGandalf (approved):

DarthGandalf:
I don't have a strong opinion on this.

Heegu-sama (approved):

PR #581, opened 2024-06-11
Unpagify intro links

DarthGandalf:
en-04


qwrtln:
Why? Those pages are fine. They just explain what it is and how it looks.

DarthGandalf:
Mm okay

PR #600, opened 2024-06-21
[HE] setup translation

davidpodless:
This is the setup of the Hebrew translation

There are few problems I couldn't solved and if someone has an idea how to fix them - it will be awesome.

1. it get stuck when running tools/build.sh he
here:
(./sections/translated/he/title_page.tex) [1<./assets/layout/tausta.png><./asse
ts/layout/bottom.png><./assets/layout/cover.jpg><./assets/layout/cover_title.pn
g><./assets/layout/cover_subtitle.png>]
! Undefined control sequence.
<argument>  \versionlabel 
               {} \input {.version} 
l.303   }
       
? 

Enter solved it, but I think it causes some graphical issues in the pdf
image

2. as you can see in the image, the English part is written backwards. it isn't a big of an issue, since it will be translated to Hebrew. The more problematic part is page numbers like this:
image

3. page 9 (and 16, 19 etc.) has green dots with numbers in it. but while the picture stay the same in English and hebrew, the dots moved. 
image

Any idea of how to solve any of the problems would be appreciated.


qwrtln:
A couple of things:
1. Please see how version is handled now in `main_*.tex` files. You're working on an outdated baseline. Fixing it will fix the problem of LaTeX compilation stopping.
2. I think graphical issues are not related to the issue above. Those images are placed on the page using coordinates. Since the direction of the text is different, coordinates will most likely need to be adjusted.
3. The dots probably suffer from the same issue. My guess is that you'll have to "mirror" the coordinates vertically. I wouldn't worry about it now, let's fix it once the setup is functional, and it's time to translate the appropriate section.
4. Please, target the PR to the `main` branch. We need to make sure all the checks are passing before merging anything. You should work on the main branch, not create a separate branch and merge once all translations are done. The risk of later conflicts is just not worth it.

README.md
```diff
@@ -106,6 +106,16 @@ Click in the table to download the most recent builds in the chosen language:
...
         </tr>
+        <tr>
+            <td>🇮🇱 עברית</td>
+            <td align="center">10%</td>
+            <td align="center">0%</td>
+            <td align="center">—</td>
+            <td align="center">—</td>
+            <td align="center">—</td>
+            <td align="center">—</td>
+            <td align="center">—</td>
```
qwrtln, line 117: Please construct appropriate URLs analogically.

tools/build.sh
```diff
@@ -8,7 +8,7 @@ esac
 LANGUAGE=$1
 
 case "${LANGUAGE}" in
-  ru|ua|cs)
+  ru|ua|cs|he)
```
qwrtln, line 11: This is good, please also add it to the workflow file in `.github/`

.github/CODEOWNERS
```diff
@@ -50,3 +50,8 @@ main_de.tex  @csagataj2 @Heegu-sama @markdermark @DarthGandalf
 **/cs.po @tomaas-zeman @Heegu-sama @qwrtln @DarthGandalf
 /translations/assets/cs/ @tomaas-zeman @Heegu-sama @qwrtln @DarthGandalf
 main_cs.tex  @tomaas-zeman @Heegu-sama @qwrtln @DarthGandalf
+#
+# Hebrew
+**/he.po @davidpodless @Heegu-sama @qwrtln @DarthGandalf
+/translations/assets/he @davidpodless @Heegu-sama @qwrtln @DarthGandalf
```
DarthGandalf, line 56:
```suggestion
/translations/assets/he/ @davidpodless @Heegu-sama @qwrtln @DarthGandalf
```

README.md
DarthGandalf: Copypaste the rows above, and change several letters of the language code

PR #605, opened 2024-06-22
Fix index misparse (and fail to build) for languages where digits are not digits

DarthGandalf:


davidpodless:
What this change does?

DarthGandalf:
When I enabled hebrew numerals, printable build started producing a bad index with something like `\dotfillיב‎` which is then parsed by LaTeX as a single command, which doesn't exist. Addition of {} makes it parse `\dotfill` correctly. It doesn't seem to hurt any other language, or even we keep 123 in Hebrew.

qwrtln (approved):

PR #613, opened 2024-06-23
[HE] player_turns

davidpodless:
That's the last section I translated before the first PR


davidpodless:
image

image

image

qwrtln:
That sole line before Necropolis note box doesn't look good. Can you do anything about it?

davidpodless:
> That sole line before Necropolis note box doesn't look good. Can you do anything about it?

The only thing I can think of is to add few \n between sections. 

let's try and see how it will look

davidpodless:
> > That sole line before Necropolis note box doesn't look good. Can you do anything about it?
> 
> The only thing I can think of is to add few \n between sections.
> 
> let's try and see how it will look

Hmmmmmm it looks like it ignored \n in the section all over the place
I need to check if that's across all sections

DarthGandalf:
I think \n is called \par here (paragraph). Or maybe some \vspace will help, no idea

davidpodless:
\\\\ is the key here

he-12
he-13
he-14

qwrtln:
Actually, it would be better to shorten the text, so that the orphaned line is not in the next column, not adding newlines to produce two orphans.

davidpodless:
Hmm 🤔 
Ok, I'll try but I fear it will break the whole structure of the page

Let's see what happens

davidpodless:
what do you think?

he-12
he-13
he-14

qwrtln:
Better!

qwrtln (approved):

PR #617, opened 2024-06-25
[HE] resources

davidpodless:
The Valuables are little higher than the rest of the resources. 😕
Suggestions what can be the causes would be welcomed

he-22


davidpodless:
Also, I know I broke the structure when I removed the 1 x valuables, but in Hebrew and would sound terrible.

davidpodless:
Here's the updated version

he-22

DarthGandalf:
I've just noticed: this should be not "x", but the \texttimes symbol (but also in original)

davidpodless:
what do you think?

he-22

DarthGandalf:
well, I don't know whether in Hebrew it's supposed to be on the left or on the right. But that's for you to decide. I'm more surprised that the location changed with the symbol change

davidpodless:
> well, I don't know whether in Hebrew it's supposed to be on the left or on the right. But that's for you to decide. I'm more surprised that the location changed with the symbol change

Do you mean the locations of the numbers? that's was on purpose, we (the WhatsApp group of the translation) and we agreed that this way it looks better

DarthGandalf (approved):

PR #634, opened 2024-07-02
Use latex way to spell latex

DarthGandalf:
en-54


qwrtln (approved):
Looks kind of... 🤨 But well...

DarthGandalf:
That's the best I could do with the image extracted from inferno guide :(

Still better than the fish eye photo we have now

tomaas-zeman (approved):

Heegu-sama (approved):

PR #637, opened 2024-07-03
[HE] units

davidpodless:
he-27
he-28
he-29


DarthGandalf:
I think we should rotate the gorgon

davidpodless:
Like that?
what do you think about behemoth? to flip or not to flip?

he-27

he-28

DarthGandalf (approved):

PR #645, opened 2024-07-20
Base/ai section updates

ohgodnotagain:
there are two info blocks in the list
first and the second, describing some steps in instructions
and we use that pattern in the whole book

make some steps:
first
second

moreover, we have "in both cases" sentence right after those two steps

so from my perspective it's quite a right change

before:
en-0

after:
en-5

after [ru]:
ru-5


sections/ai_rules.tex
```diff
@@ -101,7 +103,7 @@ \subsection*{Combat against AI}
 \vfill
 
 \centering
-\includegraphics[width=0.9\linewidth]{\art/coin.png}
+\includegraphics[width=0.7\linewidth]{\art/coin.png}
```
qwrtln, line 107: Don't resize, and please don't use different size for different languages. Just get rid of it.

Heegu-sama (approved):

PR #654, opened 2024-08-10
Add info that was omitted

qwrtln:
Added missing info about removing cards and secondary heroes using the same units. Also, adjusted the layout of combat a little, and moved some art around.

Technical:
 - bumped version to 1.3, it should be released soon
 - collapsed po files in PRs by default

Diff:
en-15
en-18
en-33
en-34

Polish:
pl-15
pl-18
pl-33
pl-34


sections/deckbuilding.tex
```diff
@@ -30,6 +30,7 @@ \subsection*{General Card Rules}
     \item \textbf{Permanent} \includesvg[height=10px]{\svgs/permanent.svg} Cards\index{Permanent Cards} stay in play until discarded or replaced.
       \textbf{You may only have one permanent Card at a time}; playing another discards the first.
   \end{itemize}
+  \item Whenever you are instructed to \textbf{remove a Card}, remove it completely from the game.
```
DarthGandalf, line 33: Perhaps mention the tournament rules here which modify this?

DarthGandalf (approved):

sections/deckbuilding.tex
qwrtln: Not sure which rules you have in mind. Do we mention them anywhere?

DarthGandalf: Выделение_610

It's in the tournament book

Heegu-sama (approved):

sections/deckbuilding.tex
qwrtln: Not sure if it belongs here. Since it's a separate book with separate rules, maybe they should stay there?

Issue #592, opened 2024-06-15
Improvements to text on AI combat


JorgenLjones:
1. Clarify that the AI will not have any walls or defensive structures if not explicitly stated in the scenario. [Confirmed by devs here](https://boardgamegeek.com/thread/3227959/article/43962848#43962848)
2. When the devs respond to [this question](https://boardgamegeek.com/thread/3227959/article/44422087#44422087), add which distance to use when determining which unit is closer.

### _Text edit suggestion specifically under AI Deck_
3. Remove the word 'automated' as it isn't any different to any other deck and could confuse the reader. 
> “AI Heroes use two automated Decks during Combat”
→
> “AI Heroes use two Decks during Combat”
4. Specify that the Spell Deck the AI will be using is separate to the regular Spell Deck.
> "the AI Deck, and the Spell Deck"
→
> "the AI Deck, and the AI Spell Deck"
5. Change 'is built using' to 'consist of' to avoid implying that one must do something special to “build” this deck.
> “The AI Deck is built using three types of AI cards: Might, Magic and Skill."
→
> “The AI Deck consists of three types of AI cards: Might, Magic, and Skill cards"
6. Clarify that the player has to ensure the correct cards are included in the AI Deck
> “Each Campaign scenario lists the number of each type of Card included in the AI Deck."
→
> “Each Campaign scenario lists the number and types of Cards to include during setup."
7. Delete the sentence “Might and Skill Cards have two variations” as it isn't essential information, and feels misplaced. Plus, Might cards have more than two variations, as not all Attack and Defence cards are identical.
8. Somewhat dependent on suggestion 7 above, simplify and be more assertive on selecting all cards at random. (Maybe even remove this sentence as well?)
> "Choose these cards randomly when building the Deck."
→
> "Shuffle the AI Deck to prepare it for setup."
9. Clarify that the scenario text will determine if Skill Cards are included. If suggestion 8 is implemented, then this sentence should come before it.
> "If Skill Cards are included, search for and set aside the Ability Card related to it."
→ 
> "If the Campaign scenario specifies an enemy Hero Ability: Find and set aside this Ability card and Include the Skill Cards to the AI Deck"
10. 
> "Build the AI’s Spell Deck by separating the indicated Spells from the regular Spell Deck.
→
> "If the Campaign scenario indicates any spells for the AI, find these Spell Cards, shuffle them and set them aside. This is now the AI Spell Deck."
[...]

PR #656, opened 2024-08-10
Fix glyphs

qwrtln:
I moved most of the glyphs slightly below the text line. It's 15% of their height for most of them, 10% for unit stars, and 0% for a handful of exceptions which did not look good: movement horse, expert crown, reinforcement arrows (used once), and levels (also used once). I tried to picture everything below.

That dangling line in AI section was driving me crazy, so I employed some of the suggestions from #592 to improve the text.

It will be a nightmare to fix in po files, I could use a hand here.

en-08
en-10
en-11
en-12
en-13
en-15
en-16
en-17
en-18
en-19
en-20
en-21
en-22
en-23
en-26
en-27
en-28
en-29
en-30
en-31
en-32
en-33
en-34
en-35
en-36
en-37
en-38
en-39
en-40
en-41
en-44
en-45
en-46
en-47
en-48
en-49
en-50
en-51
en-53


metadata.tex
```diff
@@ -105,6 +105,16 @@
...
+\newcommand{\svg}[2][10]{%
+  {\raisebox{-0.15\height}{\includesvg[height=#1px]{\svgs/\detokenize{#2}.svg}}}%
+}%
+\newcommand{\svgunit}[2][10]{%
+  {\raisebox{-0.1\height}{\includesvg[height=#1px]{\svgs/\detokenize{#2}.svg}}}%
+}%
+\newcommand{\svgeven}[2][10]{%
+  \includesvg[height=#1px]{\svgs/\detokenize{#2}.svg}%
+}%
+
```
tomaas-zeman, line 117: I'm not sure about this. You already neede to pay attention to which glyph you used and decide between two. Now it's three. 

Suggestions
- Modify glyphs so they all look the same while using the same command
- Have only `\svg` command which handles `unit` and `even` glyphs internally

qwrtln: We're not modifying the text and adding glyphs any more. If you want to simplify this, you're welcome to do this after this PR. I won't change it now that I'm 80% done. Those glyphs look much better anyway.

qwrtln: BTW, all glyphs looking the same is the issue this PR is aiming to fix.

tomaas-zeman (approved):

Heegu-sama (approved):

PR #709, opened 2024-10-07
Add test for broken cmap

qwrtln:
This is to get quick feedback if Ghostscript breaks the file's CMap again.

With Ghostscript 10.04 all the files built with LuaTeX are broken, in which case the test should fail.

### DIFF
--- a/.github/workflows/build-and-publish-rule-book.yaml
+++ b/.github/workflows/build-and-publish-rule-book.yaml
@@ -99,7 +99,11 @@ jobs:
       - name: Check for broken CMap
         if: github.event_name == 'pull_request'
         run: |
-          NEEDLE=$(grep TEST_SUBSTRING= main_${{ matrix.language }}.tex cut -f2 -d=)
+          NEEDLE="$(grep TEST_SUBSTRING= main_${{ matrix.language }}.tex cut -f2 -d=)"
+          if [[ -z "$NEEDLE" ]]; then
+            echo "No TEST_SUBSTRING found in main_${{ matrix.language }}.tex."
+            exit 1
+          fi
           pdftotext main_${{ matrix.language }}.pdf - | tr '\n' ' ' > main_${{ matrix.language }}.txt
           if grep -iq "$NEEDLE" main_${{ matrix.language }}.txt
           then
--- a/.github/workflows/build-and-publish-rule-book.yaml
+++ b/.github/workflows/build-and-publish-rule-book.yaml
@@ -10,7 +10,7 @@ on:
 env:
   PO4A_VERSION: "0.73"
   PO4A_GH_URL: https://github.com/mquinson/po4a/releases/download
-  GHOSTSCRIPT_URL: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs10030/gs_10.03.0_amd64_snap.tgz
+  GHOSTSCRIPT_URL: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs10040/gs_10.04.0_amd64_snap.tgz
   ARTIFACT_RETENTION_DAYS: 1
 
 jobs:
@@ -85,8 +85,8 @@ jobs:
           path: main_${{ matrix.language }}.pdf
           retention-days: ${{ env.ARTIFACT_RETENTION_DAYS }}
 
-      - name: Install pdfinfo
-        if: ${{ matrix.language == 'en' }}
+      - name: Install poppler
+        if: github.event_name == 'pull_request'
         run: sudo apt-get update && sudo apt-get install -y poppler-utils
 
       - name: Check number of pages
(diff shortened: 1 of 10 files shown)


.github/workflows/build-and-publish-rule-book.yaml
```diff
@@ -96,6 +96,19 @@ jobs:
           NUM_OF_PAGES=$(pdfinfo main_${{ matrix.language }}.pdf | awk '/^Pages:/ {print $2}')
           echo "NUM_OF_PAGES=${NUM_OF_PAGES}" >> "$GITHUB_OUTPUT"
 
+      - name: Check for broken CMap
+        if: github.event_name == 'pull_request'
+        run: |
+          QUOTE=$(grep -A1 heegusquote main_${{ matrix.language }}.tex | tail -1 | grep -o "{.*}" | tail -c +2 | head -c -2)
```
DarthGandalf, line 102: this tail|head deserves a comment

DarthGandalf (approved):

qwrtln:
What the heck, French? For whatever reason, the test is failing.

.github/workflows/build-and-publish-rule-book.yaml
qwrtln: Either that, or maybe you know a better way of getting rid of the {} surrounding the text? Not a fan of this tail|head.

qwrtln:
OK, French is splitting the quote into two lines
```
J’ai réalisé ce projet par amour de l’apprentissage des jeux et de la compréhension de leur
complexité.
```

DarthGandalf:
```
cat main_fr.tex | perl -E 'local $/; $x=<>; $x =~ /heegusquote.*?textit\{(.*?)\}/s; print $1'
```

qwrtln:
> ```
> cat main_fr.tex | perl -E 'local $/; $x=<>; $x =~ /heegusquote.*?textit\{(.*?)\}/s; print $1'
> ```

I had to change the line because French was having a problem again with quotes. They're rendered differently in the PDF file. This time it's without curly braces, so no need for Perl, I hope that's OK.

DarthGandalf:
> no need for Perl, I hope that's OK

It's not like I'm married to perl, or something.

Readability-wise, new version is better than before, but it seems even more fragile. It depends on the specific line which is 2 lines below the "intro" to not have any LaTeX commands in any language, right?

qwrtln:
> Readability-wise, new version is better than before, but it seems even more fragile. It depends on the specific line which is 2 lines below the "intro" to not have any LaTeX commands in any language, right?

Maybe let me state the problem first. Files built with LuaTeX and optimized with Ghostscript look good, but if you try to convert them to txt or copy any text, you get stuff like this:
```
"v h?2 *QKKmMBiv
P+iQ#2` 9- kyk9
o2`bBQM RXy
h?Bb Bb  +QKKmMBiv@/`Bp2M T`QD2+i- r?B+? ?b  :Bi>m# `2TQbBiQ`vX 1p2`vQM2 Bb r2H+QK2 iQ
+QMi`B#mi2 a+2M`BQb- KF2 +?M;2b- M/ }t 2``Q`bX
```
Since every file has its own words, I figured I need to get a piece of translated text, and check if it's present in the target file. One could argue that a simple `grep "Heegu"` would work, but I figured having 3+ words is safer.
So yes, I chose intro for this arbitrarily. Those thee lines are not obfuscated by any LaTeX commands, and the middle one contains no special characters in any language. I figured it's a good pick. I'm open to ideas if there's an obviously better way to do it.

qwrtln:
Yay, it worked for the languages built with LuaTeX. One can assume the test is effective. Now let's try with downgraded Ghostscript.

DarthGandalf:
> I'm open to ideas if there's an obviously better way to do it.

What about letting a language to specify a substring which it guarantees to exist?

e.g. add a line like this to main_en.tex:

```
% TEST_SUBSTRING=Vous êtes tous invités à apporter
```

then grep for that line

qwrtln:
> > I'm open to ideas if there's an obviously better way to do it.
> 
> What about letting a language to specify a substring which it guarantees to exist?
> 
> e.g. add a line like this to main_en.tex:
> 
> ```
> % TEST_SUBSTRING=Vous êtes tous invités à apporter
> ```
> 
> then grep for that line

1. Text from `main_*.tex` is guaranteed to exist.
2. One would need to specify it in the matrix workflow, which GitHub would visualize with every job. I hate this UI. See how it looks in the mission book PR, I'm about to change it.

DarthGandalf:
You misunderstood me.

I said, add that string to main.tex, literally, as a comment. Then in the workflow grep for TEST_STRING (literally), then grep for its result in the pdf

DarthGandalf:
> Text from main_*.tex is guaranteed to exist.

Sure. But "the text exactly 2 lines below 'intro' string" isn't.

qwrtln:
> I said, add that string to main.tex, literally, as a comment. Then in the workflow grep for TEST_STRING (literally), then grep for its result in the pdf

OK, should we repurpose the intro line for this?

DarthGandalf:
No, because you can't guarantee the intro text will have the same arbitrary property in all languages, which happened to be true now.

tomaas-zeman (approved):

PR #711, opened 2024-10-09
[HE] credits

davidpodless:
I don't think it is right to try and write the names in Hebrew letters.

he-52

### DIFF
--- a/translations/credits.tex/he.po
+++ b/translations/credits.tex/he.po
@@ -75,7 +75,7 @@ msgstr ""
 #. type: Plain text
 #: sections/credits.tex:4
 msgid "\\addsection{Credits}{\\images/experience.png}"
-msgstr ""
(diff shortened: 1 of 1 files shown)


qwrtln (approved):
Suit yourself, LGTM.

qwrtln:
Oh, BTW, there's a text regarding GitHub in `main_he.tex`, mind translating that too?

davidpodless:
> Oh, BTW, there's a text regarding GitHub in `main_he.tex`, mind translating that too?

I must have missed it... I'll do it now

DarthGandalf:
The hyphenation (Bar-tosz) and newlines inside names look weird to me. Would it look any better to disallow hyphenation here, and make the space between Name Surname to be non-breakable?

I don't speak Hebrew though, maybe this is good as is

PR #722, opened 2024-10-23
[DE] Translation Map-Elements

DerSchlupf:
German Translations, but still with english cards:
image
image


DerSchlupf:
Hey people,
in the starting you find the "old version" and the new one (with headline "compare with"). I have short the text but the layout removed the free line at the top under the header "Kategorien der Orte" instead of moving the text more up and creating a free line at the end of the text. I was hopefully, that this one line would be enough to put the picture under the two texts.

qwrtln:
> Hey people, in the starting you find the "old version" and the new one (with headline "compare with"). I have short the text but the layout removed the free line at the top under the header "Kategorien der Orte" instead of moving the text more up and creating a free line at the end of the text. I was hopefully, that this one line would be enough to put the picture under the two texts.

LaTeX does that because it's trying to keep both columns the same size. I made some hacks to fit everything on a single page. It's very tight, though.

DerSchlupf:
Any suggestions? If not:  I've done the final review and This Page is also ready for merge. As i said in another Thread i will focus on 100% texttranslation first, so its no problem for now, that the page contains english images.

qwrtln (approved):

PR #739, opened 2024-11-05
[HE] translate unit box and fix tables

davidpodless:
he-29
he-40

now the tables are Right to Left, which make more sense. I also translated the last box in units

Only index left :-D


qwrtln (approved):

davidpodless:
> It feels like the last row should be more narrow now, since it's only 1 line there

There are a lot of boxes like this in the translation. I don't like the idea to change the layout
I can try to make it 2 lines if you prefer

DarthGandalf:
Not a too big deal

PR #740, opened 2024-11-05
Make back cover tables pixel perfect

qwrtln:
This is part 1 of 3. Combat and Units will follow after DE and HE translation PRs are merged. Pay attention to the right margin of the tables.

Before:
image

After:
image

Some customizations needed to be made for Hebrew (naturally), but also Russian (Уровень VII was too wide) and Ukrainian (one header in the trade table) because the cells were too narrow to accommodate the text. See the images below. I made some adjustments, so the end result is the same.

Russian (the leftmost cell):
image

Ukrainian (the header with gold):
image


Heegu-sama (approved):

DarthGandalf:
The PR description is confusing: I see the difference at the right margin, not at left margin. But I like the change, thanks

PR #741, opened 2024-11-05
Make difficulty table margins equal

qwrtln:
Only German and Ukrainian required adjustments, all the other languages looked good. See the side margins of the original, which are equal now:

en-5


DarthGandalf:
How did you come up with the new numbers?

How to do the numbers for new tables if we add any?

DarthGandalf (approved):

qwrtln:
The answer to both questions is: trial and error.

DerSchlupf:
The 4th Variante needs two lines in the german Version but that's the only difference, isnt it? And because of this the devil is smaller, right. It could be difficult to short this line more. It was also Tricky in the previous table Version. But i will search for shorter wordings in the future.

qwrtln:
Yes, but it still fits on the page, so no problem.

Heegu-sama (approved):

PR #744, opened 2024-11-07
Make medusa art smoother

qwrtln:
How does it compare on the page? Worth the change?

before|after
en-42


Heegu-sama (approved):
She looks fine. Slightly blurry at spots like the waist but better overall.

qwrtln:
> She looks fine. Slightly blurry at spots like the waist but better overall.

AI - can't live with it, can't live without it.

PR #749, opened 2024-11-12
New SP clarification

DarthGandalf:
en-20


DerSchlupf:
Nice Addition IMO!

DerSchlupf (approved):

qwrtln:
Does it have to be a box? Having it as a last sentence of the paragraph would be clearer, shorter, and wouldn't disrupt the flow of text:
image

DarthGandalf:
@qwrtln True. Could you push that?

I made it the box because such rule would be relevant very rarely, making it into Exceptions category

qwrtln:
I don't consider it an exception. I guess it's implicitly obvious to most people, but it should be enough for those complaining that they can't get gold from View Air, if they have Air Magic on the table.

qwrtln (approved):

Heegu-sama (approved):
Nice clarification. I assume this was confirmed on discord/BGG at some point.

DarthGandalf:
Yep, https://discord.com/channels/740870068178649108/1211591995328299038/1305812432543023146

qwrtln:
Was merging this necessary? Now every translation is fuzzy.

DarthGandalf:
Would you prefer to update all the translations before merging?

qwrtln:
No, release 1.3 first. We had 7 languages fit for release. Now we have 1.

DarthGandalf:
Hm, right.

let me prepare rollback then

qwrtln:
Not sure why people are so attached to releases, but I keep seeing them using the old 1.2 PDF.

PR #752, opened 2024-11-14
[DE] Update Gamemodes

DerSchlupf:
updating game-mode section
image


DerSchlupf:
Because of some space in this section i've added a short note, how neutral units are controlled in conflict, alliance (controlled by enemy faction) and coop(controlled by AI). In the original rewrite there is this note only in solo-campaign (controlled by AI). do you like it? i think it would be a helpful addition also for the original?

DerSchlupf:
Oh review is done. If you are fine with my additions its ready for merge.

qwrtln:
You shouldn't make up your own text like that. It's better if you add it to the original.

DerSchlupf:
Yeah you are right and i agree. I will delete it in the next upload, but want to see how it looks like :)

But what do you think about it? It is worth to implement in the original? IMO such a wise you have a better over view about specific changes the chosen game Mode has. But what do you think?

qwrtln (approved):

PR #753, opened 2024-11-14
[DE] Update Intro

DerSchlupf:
updating intro session:
image


DerSchlupf:
po4a commit

DerSchlupf:
The original sentence about using capital letters makes no sense in german, because too much words are regulary written in large letters (e. g. every noun) and nobody could differ between normal words and gameplay-relevant-words only because of large letters. i've deleted it and added a \par. other than that similiar than before and IMO ready for merge.

qwrtln (approved):

DarthGandalf (approved):

PR #754, opened 2024-11-14
Revise scripts in tools/

qwrtln:


tools/build.sh
```diff
@@ -1,13 +1,83 @@
...
+    echo "Error: Invalid language code '$1'. Valid codes are: ${valid_languages[*]}" >&2
+    exit 1
+  fi
+fi
+
+# Parse remaining command line arguments
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    -p|--printable)
+        printable=true
```
DarthGandalf, line 53: it'd probably be simpler to avoid these extra variables, and export the env var right here

tools/build.sh
```diff
@@ -1,13 +1,83 @@
...
+  echo "  -p, --printable    Enable printable mode"
+  echo "  -n, --no-bg        Disable background"
+  exit 1
+}
+
+# Function to check if language is valid
+is_valid_language() {
+  local lang="$1"
+  for valid_lang in "${valid_languages[@]}"; do
+    if [ "$lang" = "$valid_lang" ]; then
```
DarthGandalf, line 31: [[ ]] pls

Heegu-sama (approved):

tools/build.sh
qwrtln: done

DarthGandalf (approved):

PR #755, opened 2024-11-15
Create release pipeline

qwrtln:
This is a workflow of convenience, which makes building release files easy. The PR also updates the optimize script with experimental monochrome build.

See the [workflow result](https://github.com/Heegu-sama/Homm3BG/actions/runs/11857453520).

### DIFF
--- a/tools/optimize.sh
+++ b/tools/optimize.sh
@@ -1,17 +1,79 @@
 #!/usr/bin/env bash
 
-LANGUAGE=$1
-CMYK=$2
+# Default values
+LANGUAGE="en"
+CONVERSION_MODE=""
 
-if [[ ${CMYK} == "cmyk" ]]
-then
-  ARGS=-sColorConversionStrategy=CMYK
+# Valid language codes
+VALID_LANGUAGES=("en" "pl" "es" "fr" "ua" "ru" "cs" "he" "de")
+
+# Function to print usage information
+usage() {
+  echo "Usage: $0 [language] [--mono|--cmyk]"
+  echo "Positional arguments:"
+  echo "  language           Language code: ${VALID_LANGUAGES[*]}"
+  echo "                     Defaults to 'en' if not specified"
+  echo "Options:"
+  echo "  --mono             Convert to monochrome (black and white)"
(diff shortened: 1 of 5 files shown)


DarthGandalf:
What's the point of monochrome build? When printing the colorful page on a b/w printer, it'll automatically become monochrome

tools/optimize.sh
```diff
@@ -1,17 +1,79 @@
...
+# Optimize the PDF
 gs -o main_${LANGUAGE}_optimized.pdf \
-  -sDEVICE=pdfwrite \
-  -dCompatibilityLevel=1.5 \
-  -dPDFSETTINGS=/prepress \
-  -dDetectDuplicateImages=true \
-  ${ARGS} main_${LANGUAGE}.pdf
+   -sDEVICE=pdfwrite \
+   -dCompatibilityLevel=1.5 \
+   -dPDFSETTINGS=/prepress \
```
ExeVirus, line 77: I would consider making this pipeline longer and include the output for a /screen optimized output. I would expect a significant file-size reduction that would be valuable for those of us who do not print the rules and want to pull it up fast on a mobile device or tablet. Further, if you are going to have the BW option, should consider a 3rd most compressed version:

- Heroes3_English_Rules_Rewrite_1_3_printable
- Heroes3_English_Rules_Rewrite_1_3_small
- Heroes3_English_Rules_Rewrite_1_3_small_bw

ExeVirus: Because - keep in mind that 99% of users won't pull the pipeline, they will come here and grab the latest release downloads from the pipeline artifacts

qwrtln:
I'm not concerned with the file size. The monochrome build saves only ~4 MB. I was thinking that the file having a proper monochrome color profile might produce better print quality, then direct conversion by the printer. I don't have any data to support that claim, though. It's possible that letting the printer handle that is better. I'll remove the conversion from the workflow.

tools/optimize.sh
qwrtln: It's actually 100% ;-) No user will run it, that's not the point here.

Heegu-sama (approved):

PR #760, opened 2024-11-25
Replace images with LaTeX circles

DarthGandalf:
This will help with translations, because they only need to replace the card image with the localized image.

en-19
en-20
en-38
en-28
en-07


qwrtln:
Oh boy, this is going to be tough on translators. Luckily, only the French crew is left to fix it. In any case, good step forward.

pl-07
pl-19
pl-20
pl-28
pl-38

qwrtln:
@Gastonia02 can you fix it for French?

DarthGandalf:
> Oh boy, this is going to be tough on translators.

This is somewhat tedious, but it's actually easier for me, because updating all the translations follows the same pattern. Except for French because I don't want to add circles on top of circles. And for Hebrew because I don't dare touching it due to being confused by interaction between LTR and RTL.

DarthGandalf:
This is good enough for now. I didn't reconstruct example images yet, only cards. Maybe later.

qwrtln (approved):

Heegu-sama (approved):

PR #762, opened 2024-11-26
[DE] Translation AI-Rules

DerSchlupf:
Final Version.
image

Printable:
image


DerSchlupf:
Some untranslated phrases, perhaps something to rename and to short.

DerSchlupf:
Why doesnt Latex move the text in an order like that:
image

Then I could shorten the text a bit and would almost have a column free for the titan,  or, if in doubt, make the titan slightly smaller, that it fits.

DarthGandalf:
Normally, it would. But there's an explicit `\vfill` just before AI decks header, which fills the remainder of the page with empty space.

translations/ai_rules.tex/de.po
```diff
@@ -88,6 +103,15 @@ msgid ""
...
+"Bot-Helden können nicht kapitulieren und du kannst vor ihnen nicht kapitulieren;\n"
+"sie werden immer kämpfen, bis sie keine Einheiten mehr haben.\n"
+"Wenn man einen Kampf gegen einen Bot-Helden gewinnt, erhält man keine Belohnungen, außer dies ist im Szenario angegeben.\n"
+"Bot-Helden haben weder ein Stadt- oder Heldentableau noch Ressourcen.\n"
+"Ihre Einheiten sind statisch und werden durch den Aufbau des Szenarios oder durch andere Regeln definiert.\\par\n"
+"Alle Abweichungen von den oben genannten Regeln werden in den Regeln des jeweiligen Szenarios beschrieben.\n"
+"\n"
 
 #. type: multicols*
 #: sections/ai_rules.tex:32 sections/ai_rules.tex:115 sections/combat.tex:185
```
DarthGandalf, line 117: It's the one just below here.

po format is designed to dedup strings, so that the same original string is translated to the same string. But with the way how we use it with po4a, it makes some of these strings which are actually formatting, hard to "translate" without affecting other parts of the book

DarthGandalf:
You can try to remove \vfill from sections/ai_rules.tex and if that works any better for you, we can try to reconcile how that paragraph works across translations

DerSchlupf:
If i would try, is it right that i make a simple git push in this PR with two changed files: the changed tex-file and po-file?

DarthGandalf:
Probably yes. po4a check will complain, but for the quick test that's fine.

Or build it locally instead of pushing

DerSchlupf:
po4a commit

DerSchlupf:
Wow, it worked nearly completly succesful. :D I havent checked, if other languages are build normally because of the changed tex.file. but german looks good. i still have to short some lines but i think that should be possible. 

Whats your opinion about this small test? And what to do with this tex-file? Tell me, if i have to make it undone and should search other solutions.

qwrtln:
Instead of changing it for every language, we could glue this `\vfill` to the previous paragraph. That way, we won't have to alter it anywhere else.

sections/ai_rules.tex
```diff
@@ -28,8 +28,6 @@ \subsection*{AI Hero Turn}
 Their Units are static and defined by the Scenario's setup or other rules.\par
 Any differences to the above will be described in any given Scenario's own rules.
```
qwrtln, line 30:
```suggestion
Any differences to the above will be described in any given Scenario's own rules.
\vfill
```

qwrtln: Do this, and then `po4a commit`.

DerSchlupf:
po4a commit

translations/ai_rules.tex/de.po
```diff
@@ -172,36 +232,50 @@ msgid ""
...
-#: sections/ai_rules.tex:88
+#: sections/ai_rules.tex:87
 #, no-wrap
 msgid ""
 "\\begin{itemize}\n"
 "  \\item Enemy Ground \\svg{unit_ground} and Flying \\svg{unit_flying} Units prioritize attacking Units of the \\textbf{same} tier.\n"
 "  If this is impossible, they attack the Unit of a lower tier (in tier \\textbf{descending} order, down to bronze), and if that is also impossible, they attack the Unit of a higher tier (in tier \\textbf{ascending} order).\\par\n"
 "\n"
 msgstr ""
+"\\begin{itemize}\n"
```
qwrtln, line 272:
```suggestion
"\\begin{itemize}[leftmargin=15pt]\n"
```

qwrtln: This, plus get rid of this orphaned "ab", and everything fits nicely.
image

qwrtln:
Other languages will need fixing, though.

DerSchlupf:
I've build the pdf and it seems to be the old problem. Where is the mistake?
image

DerSchlupf:
oh i missed your answer before. I will do it

DerSchlupf:
Okay, final version in the first post with the first german card. Any suggestions? If not i think its ready for merge.
But i dont know, which consequenzes the changed tex-file has for other languages.

qwrtln:
Please reduce the file size. The original graphic has ~400KB, your file has almost 6MB. There's no need for it to be so big.

DerSchlupf:
oh, i think i have to ckeck this for all other graphics...^^

DerSchlupf:
Anything to do here? If not it is ready for merge.

qwrtln:
Would it be possible to get rid of "ab" and "fordert"?

image

DerSchlupf:
Done! Thanks for your suggestion. It looks better now.

Heegu-sama (approved):

PR #777, opened 2025-01-02
Move schools of magic from exp content

DarthGandalf:
Also clarified further than Specialties are not spells

Noticed that spells and war machines use different font in legend than other legends, fixed that

en-20
en-21
en-22

en-44
en-45


DarthGandalf:
In almost every expansion *some* effect uses this. Even the tournament book refers to basic vs expert spells, and that's in the core box.

Also this strengthens the point about hero specialties' similarity to spells, because they also use schools.

Heegu-sama:
> What is the point of this? Schools of Magic subsection doesn't even mention the cards. Simply moving it makes little sense.

I think it makes a lot of sense to have the schools mentioned here already. They are very prevalent in any given expansion and I doubt anyone just buying the core game would be reading this.

Heegu-sama (approved):

DerSchlupf:
I've see, that german page-check fails. Is there something to do before merging? Or should i simply have to wait and open a new PR with german corrections?

DarthGandalf:
@DerSchlupf there are several options:

* you can push the update to this branch directly. That's what I'll do with Russian before merge
* you can push the update to another branch, and open PR from that branch to this branch
* you can wait until I find a temporary workaround how to make the build not fail, and fix it later in the main branch as usual. Since you're active, I'd prefer to not add such workarounds

qwrtln:
> I think it makes a lot of sense to have the schools mentioned here already. They are very prevalent in any given expansion and I doubt anyone just buying the core game would be reading this.

My concern was around mentioning them verbatim from the other section, not moving. This is a section about deck building, and those cards are not mentioned whatsoever. Maybe it's irrelevant here, I don't know.

In any case, I'd suggest replacing the dwarf (BTW, how did you make it so pixelated? There's a good quality image available), which doesn't fill up the page. Efreet from page 7 of the battlefield rule book would fit just fine, I think. Also, please increase the spacing between the war machine card and the rest of the text.

Page 21 looks very crammed right now. There's only so much text and art you can fit into 56-page long booklet. If you want to keep expanding, you'll eventually need more pages. I liked the gameplay example on its own page.

img-109

DerSchlupf:
po4a commit

DarthGandalf:
> My concern was around mentioning them verbatim from the other section, not moving. This is a section about deck building, and those cards are not mentioned whatsoever. Maybe it's irrelevant here, I don't know.

Sorry, can you rephrase?

I also reworded this subsection, not just moved text from A to B. This whole chapter is about deck, this section is about spell cards, and schools of magic relate to spell cards. Which cards are not mentioned where?

> BTW, how did you make it so pixelated?

I now wonder the same. Just curious: any specific reason you prefer efreet over dwarf?

I chose dwarf because I had this file in assets/ and it wasn't use in other pages. But apparently it wasn't even in git, so I don't know where I got it from.

> Page 21 looks very crammed right now. There's only so much text and art you can fit into 56-page long booklet. If you want to keep expanding, you'll eventually need more pages. I liked the gameplay example on its own page.

That's true. The difficulty is that increasing page number by 1 in the middle cause next pages flip their oddity, which I tried to avoid. And increasing the number of pages by 2 requires more content. I could move war machines and enhanced stats here, because both are the cards in your deck, but Heegu doesn't like that.

I can try to mitigate flipping oddity in other ways, like moving the extra quote page to different place, or mirroring the gorgon to avoid this gap:

Выделение_649

DarthGandalf:
@DerSchlupf before we finalized how English layout should look like, I wouldn't bother updating translations, that's just extra work which will likely go to waste...

DerSchlupf:
Okay in german there is still something to do. But the general translations are done. Probably you found solutions without shorting text. If not, ping me again. 

Deckbuilding: 
image

Expansion content:
image

qwrtln:
> Just curious: any specific reason you prefer efreet over dwarf?

> I'd suggest replacing the dwarf [...], which doesn't fill up the page.

Just read the comment.

qwrtln:
> Ah, I see. It wasn't clear that "doesn't fill up the page" is the reason.

Don't worry, reading comprehension gets easier with practice.

DarthGandalf:
> Don't worry, reading comprehension gets easier with practice.

That it does...

@qwrtln PTAL.

P.S. while tinkering with how to unbreak the gorgone, I accidentally added some controls to show/hide Notes page and whether to show the quote on a separate page

qwrtln:
Efreet is an improvement. Good choice of putting it in the left column. Although if you follow through with the suggestions below, you might want to put it in the right one. Not sure now.

To improve the flow of text with all the cards, I suggest putting the art in the left column, and replacing the ugly azure dragon (the artistic quality is noticeably lower) with something else (e.g., rust dragon, page 4 in the fortress mission book). Something like this (the way it's intended to be read in print):

drag

Please swap the fire magic and air magic transparent art on the render you've shown. Fire magic plays really well with the example.

Also, I'd rather get rid of the quote page for now instead of notes. The mostly empty quotes page position was chosen deliberately to make a series of following two-page sections appear together as an open book. AI rules, difficulty and expansion content are "broken" in this case. Either remove the page or move if just before all map locations.

The gorgon needs additional hspace for the printable build only (plus probably a slightly bigger size to make up for the movement).

Finally, if you decide to remove the notes page, I think index and credits should be swapped. This way, the index will be easier to access on the inner page of the back cover.

DarthGandalf:
The main reason I chose azure dragon there is because its blue color matches the blue bottom of the specialty card shown :D but sure, rust dragon is fine. And the reason why I've put it in that column is to make the specialty visually further away from spells which it's often confused with. Doesn't really make much difference.

Even with notes shown, I think it makes more sense to have "notes > credits > index" or "credits > notes > index"

All done

metadata.tex
```diff
@@ -76,6 +76,21 @@
 \newtoggle{noartbackground}
 \newtoggle{githubbuild}
 
+% Control addition of certain pages in the printable version.
+% Useful to make the printable version have correct number of pages.
+\newtoggle{printable_quote_separate_page}
+\newtoggle{printable_notes_page}
+\togglefalse{printable_quote_separate_page}
+\toggletrue{printable_notes_page}
```
qwrtln, line 84: I hope it's not going to get out of hand.

qwrtln (approved):
IMO, this looks much better. The azure dragon made sense when it comes to colors, but just compare it to all the faction art. You can see it's been rushed and made a lot worse.

sections/deckbuilding.tex
```diff
@@ -173,54 +171,102 @@ \subsection*{\pagetarget{spells}{Spell Cards}}
...
 
-\note{6}{
-  When you use an effect which lets you cast a Spell from somewhere other than your hand (such as directly from the Spell Deck), place that Spell into your Discard Pile after casting it.
-}
+\columnbreak
+When you use an effect which lets you cast a Spell from somewhere other than your hand (such as directly from the Spell Deck), place that Spell into your Discard Pile after casting it.
+
+\begin{center}
+  \vspace*{\fill}
+  \includegraphics[width=0.9\linewidth]{\art/air_magic.png}
```
qwrtln, line 205: I'm introducing transparency package in the other PR. I think it's better to have the art non-transparent, and regulate the alpha channel with LaTeX moving forward.

metadata.tex
DarthGandalf: Me too

sections/deckbuilding.tex
DarthGandalf: I agree

DarthGandalf:
> Okay in german there is still something to do. But the general translations are done. Probably you found solutions without shorting text.

Shortening wasn't necessary since the new layout, except for issues like gorgone which aren't directly related to translations. But I didn't try to plug your translations back in, as I've noticed that it partially undoes what I did. I've resplit the translation of searching example though

DerSchlupf:
I've seen in the builded german PDF that several Parts (even untouched Parts like "Event cards") are translated back to English. What to do from my side?
Should i pull this branch and push the missing translations into this branch? Or open a New PR? i asked thus before, but now im not sure because this PR has merged, hasnt it? Is it possible to add my changes to this already merged branch?

DarthGandalf:
> translated back to English

That's what fuzzy in .po does. When the source string is slightly changed, the program which writes .pot/.po files (po4a in our case) find similar string in the previous version, and declares that it was the string which changed, to provide easy to access old version of translation. Which needs to be updated to the current text.

> Or open a New PR?

this

The main reason (before the merge) to push the translation to the same branch is to avoid adding workarounds for the case if the mix of untranslated English doesn't fit the space. But that didn't happen with the new layout, so I decided can just merge now. All the fitting issues were random (probably due to change of odd/even pages, not sure) and not actually related to the text translated.

Other than that, it doesn't really matter whether the translation is merged to main together with English text, or after.

DarthGandalf:
In event cards specifically, they changed the formatting of the legend - font size and boldness, to match other legends

PR #211, opened 2024-04-21
Scenario end: solo campaign

DarthGandalf:
both-5
both-39

PR #778, opened 2025-01-02
Mention Tournament and Battlefield in game modes

DarthGandalf:
en-05
en-39


qwrtln:
Never played tournament, but is it an actual separate scenario type? I thought those were just additional rules for existing (clash?) scenarios. Also, Skirmish and Adventure modes are not scenarios. You could mention that those are additional game modes introduced by the Battlefield Expansion, but don't include them in the scenario list.

DarthGandalf:
It is a separate scenario type. It is pretty similar to Clash (unlike e.g. Skirmish or Alliance), so I suppose it can be considered a submode of Clash. But because of the way how map is constructed you can't just take any Clash scenario and declare that it's now in tournament mode.

The Tournament Book lists 3 such scenarios

Re Skirmish, yeah, they are not scenarios. Not sure how to represent that though

DerSchlupf:
Hi,
In general i like your suggestions to add Infos about these other Modes. Aditionally i could imagine to suggest some more changes to this page. What is the best way to show it? Can I pull this branch (if Yes, how?), then edit it for myself locally and Show a Screenshot? Or should i give you my suggestions here via Github? 

Tomorrow i have the time to tell you more Details.

DarthGandalf:
`git fetch; git checkout modes` should do it

DerSchlupf:
Okay here some suggestions. 

1) I've moved the general infos about all scenario-types in the introduction-part (playercount; scenario-endings). 
2) I've added all missing general differences between scenario-types (controlling units; trading). Something forgot?
3)  Regarding this Part in the combat section: Probably it would be nice to move the text for campaing and cooperative in the AI-Rules-Section and write in the combat section just something like: "In campaign and coop play you have to follow the "AI-Rules" (Link) for Combat Setup." 
image

And here is the page. the link under "controls" leads to the main page of Combat, but should actually only lead to the relevant part (see screenshot above), but I don't know how to do that. And I havent spend time to shorting text. But probably it isnt a problem to short the hellebardier a bit. There is much space on this page. 

What do you think?

image

DarthGandalf:
Trading... maybe. But why during selection of the mode would a player care who controls neutral units?

Since this is very start of the book, the player isn't expected to know all the rules yet while reading this. So the shorter the summary of the mode is, the better. We shouldn't dump all the corner cases onto the player from the start.

qwrtln:
My suggestion would be to add another paragraph after tournament which mentions battlefield expansion explicitly, and then adventure and skirmish separately. The former replaces map exploration and town with adventure cards, and ends with a battle on the big battlefield. The latter skips the adventure, and you jump straight to the battle. I think that's enough info. Merging two modes in a single header is a little confusing.

DerSchlupf:
> Trading... maybe. But why during selection of the mode would a player care who controls neutral units?
> 
> Since this is very start of the book, the player isn't expected to know all the rules yet while reading this. So the shorter the summary of the mode is, the better. We shouldn't dump all the corner cases onto the player from the start.

Hmm probably too detailed... I dont know^^ 

It is by the same reason, why you announce, that in campaign mode Ai Rules apply. The players get in impression how different combats work in this game. And It is a great strength of the game, that neutrals in clash/alliance are controlled by a **human** enemy player (and not by a mechanical automatism). everybody knows than, that there is less downtime in combats and so on. 

But words like "on your right" is really not necessary, i agree.

DarthGandalf:
@qwrtln Something like this?

en-05

qwrtln:
Yes, the text looks good, but the layout could use some improvement. I think the archer outlived his purpose. I have an idea.

qwrtln:
Something like this. Clear visual separation fitting the entirety of the page.

en-5

DarthGandalf:
That looks great! please push

qwrtln:
Done with a slight rewording.

qwrtln:
po4a check

sections/game_modes.tex
```diff
@@ -22,17 +22,25 @@ \subsection*{Cooperative}
...
+\begin{center}
+  \transparent{0.3}{\includegraphics[width=\linewidth, keepaspectratio]{\art/black_market.png}}
+\end{center}
 
-\subsubsection*{Adventure}
-The map exploration is replaced with Adventure cards. It ends with a battle on a big battlefield.
+\columnbreak
 
-\includegraphics[width=\linewidth]{\art/archer.jpg}
+The \textbf{Battlefield Expansion} adds two scenario-free 1 vs.~1 modes:
```
DarthGandalf, line 34: I would agree with this rewording if the rewritten book provided the rule for these. But it doesn't yet

sections/game_modes.tex
```diff
@@ -22,17 +22,25 @@ \subsection*{Cooperative}
...
 
-\subsubsection*{Skirmish}
-Just skip the adventure, and jump straight into the battle.
+\subsection*{Adventure}
+Explore using Adventure Cards instead of the Map, culminating in a hex-battlefield battle.
```
DarthGandalf, line 37: "Explore using Map" sounds wrong

sections/game_modes.tex
```diff
@@ -22,17 +22,25 @@ \subsection*{Cooperative}
...
+
+\subsection*{Skirmish}
+Direct combat without exploration.
```
DarthGandalf, line 40: I think this is too short. Are you saying the combat can be direct and indirect?

sections/game_modes.tex
qwrtln: Make it as you prefer, my main intention was to improve the layout.

DarthGandalf:
Unfortunately, gnoll is too tall (and in the picture you posted above it's thinner than the text, that's why it fits there), and the basilisk is too short, but crusader does

sections/game_modes.tex
```diff
@@ -21,10 +21,26 @@ \subsection*{Alliance}
...
+  \transparent{0.3}{\includegraphics[width=\linewidth, keepaspectratio]{\art/black_market.png}}
+\end{center}
 
 \columnbreak
 
-\includegraphics[width=\linewidth]{\art/archer.jpg}
+In addition, the \textbf{Battlefield Expansion} introduces two 1 vs.~1 modes which don't use Scenarios (see the expansion's rulebook):
+
+\subsubsection*{Adventure}
+The Map exploration is replaced with Adventure Cards. It culminates in a battle on a big hexagonal battlefield.
```
qwrtln, line 36:
```suggestion
The Map exploration is replaced with Adventure Cards, culminating in a tactical combat on a large hexagonal battlefield.
```
How about this? I think tactics is the main selling point for this mode.

sections/game_modes.tex
```diff
@@ -21,10 +21,26 @@ \subsection*{Alliance}
...
+
+\subsubsection*{Skirmish}
+Skip the adventure, and jump straight into the battle.
```
qwrtln, line 39:
```suggestion
Skip the exploration phase and engage directly in tactical combat.
```
Also, this. I just don't like "jumping" in this context.

sections/game_modes.tex
```diff
@@ -21,10 +21,26 @@ \subsection*{Alliance}
...
+A 1 vs.~1 mode with different way of building Map and scoring. See the Tournament Book.
+
+\begin{center}
+  \transparent{0.3}{\includegraphics[width=\linewidth, keepaspectratio]{\art/black_market.png}}
+\end{center}
 
 \columnbreak
 
-\includegraphics[width=\linewidth]{\art/archer.jpg}
+In addition, the \textbf{Battlefield Expansion} introduces two 1 vs.~1 modes which don't use Scenarios (see the expansion's rulebook):
```
qwrtln, line 33:
```suggestion
In addition, the \textbf{Battlefield Expansion} introduces two focused 1 vs.~1 modes without Scenarios (see the expansion's rulebook):
```
And here I think it flows better without the relative clause.

qwrtln (approved):

DerSchlupf:
On last question/suggestion to this: i understand, that it is probably to detailed to add trading and so on to the game Modes.

But what about moving some General Infos/links in the introduction Part?

In my opionion it seems weard to read that clash is only for 1-3 players (because it depends on using expansions, not on playing Clash Mode). And linked pages like "scenario end" are important for All Modes, not only campaign.

DarthGandalf:
@DerSchlupf In general I agree (as long as amount of details is not too big). But I don't know how to phrase this better.

Having the special mention of scenario end in the campaign mode was kinda my fault in #211 but yes, not only campaign mode has special rules about scenario end.

Clash and Coop are limited to 3 because they require scenarios, and (at the time of writing that a while ago) no official scenarios for 4+ players existed for these modes. Currently there's an unofficial mission book

DerSchlupf:
@DarthGandalf
Sounds good. Tomorrow morning i would give it a try to Phrase it a little different and Show you my solution (without many Details). But it would be nö Problem to prefer the current one.

Heegu-sama (approved):
Looking good.

DerSchlupf:
Hi together, 
I experimented a bit and to get this following layout i had to drop the black-market-art and i had to add some more details like trading... sorry for that :D 

I don't think it's overloaded with too many details, but you decide whether there's anything useful in there.

@DarthGandalf : Please tell me, if you have the final layout. I would do german translation before merging. I think i have the time to do it soon. 

Here the page:
image

DarthGandalf:
@DerSchlupf looks good to me, except for -rules having the hyphen. "As well" in coop may be confusing, because in fact the trading rules for coop and alliance are different.

Heegu?

Heegu-sama:
I think it's fine for now (besides those micro changes you mentioned). If there's anything else I'd personally change I'll do so once I do a full proofreading again (I promise I'll do one this year 🤣).

DerSchlupf:
po4a commit

DerSchlupf:
Hi, 

I've pushed the above version (see screenshot below). I've removed the hyphens and replaced "as well" with "also". But please feel free to reword it as you like. I'm also not sure, if there are more words which should be written in capital letters, because it is mentioned in the introduction page, that the game relevant wordings are written in capital letters.

image

sections/game_modes.tex
```diff
@@ -3,31 +3,28 @@
...
-Each game of Heroes III is played using a Scenario from the Mission Book.
-There are several types of Scenarios:
+Each game of Heroes III is played using a \textbf{Scenario} from the Mission Book. 
+The core game only allows you to choose scenarios for 1-3 players. For 4 or more players, expansions are required.\par
+In addition to the scenario objectives, read the \pagelink{Scenario end}{general victory conditions} depending on the scenario type.\par 
+There are several types of scenarios:
 
 \subsection*{Clash}
-A fully competitive mode for 2--3 players.
+A fully competitive mode. Your opponents even control neutral units in your combat.
```
qwrtln, line 12:
```suggestion
A fully competitive mode.
Your opponents control Neutral Units in your Combat.
```

sections/trading.tex
```diff
@@ -8,14 +8,14 @@
 The \pagelink{Trading Post}{Trading Post Field} and other effects allow you to either:
 \begin{itemize}
   \item \pagetarget{Trading}{trade}\index{Trading} multiple Resources with the game in accordance to the \hyperlink{Trade Table}{table at the back cover},
-  \item \textbf{remove a card} from your hand at the trading post to gain 1 \svg{gold},
+  \item \textbf{remove one card} from your hand at the trading post to gain 1 \svg{gold},
```
qwrtln, line 11: Please contain it in the other PR.

DarthGandalf:
> And here is the page. the link under "controls" leads to the main page of Combat, but should actually only lead to the relevant part (see screenshot above), but I don't know how to do that.

Done. Though differently, by adding a sentence to AI section.

DarthGandalf:
@DerSchlupf need your help with fixing German translation. The new sentence I added to AI rules doesn't fit, I temporary replaced it with TODO in German

DerSchlupf:
I've done the german translation without failing checks, yeah :D A few fixes are to do.

image
image

PR #795, opened 2025-03-31
Add suggestions from issues: spells, paralysis

### LINKED ISSUE #538: Paralysis Token description
Currently Paralysis Token description is as below:

Some effects place a Paralysis Token on Units.
That Unit must skip its next Activation. 
Remove the Token instead of activating it.
If the Unit is attacked or takes any damage
before that time, remove the Token. 
The Token does not prevent Units from 
performing Retaliation Attacks.

I think it could be put in a much better way because:
1. "That Unit **must skip its next Activation**. " doesn't really fit the situation. It does not exclude that the placement of the token is responsible for skipping the unit's activation and it suggests that skipping is unavoidable.
2. Text in general does not imply that the true impact of the token is at the unit's activation.

Is it possible to comprehend how the token works? Probably yes, but I think the description can still raise a few questions and ambiguities.
Here's my proposal then:

Some effects place a Paralysis Token on Units.
If a Unit would Activate with a Paralysis Token on it, 
you skip its Activation and remove the token instead.
If the Unit is attacked or takes any damage
while having the Paralysis Token, remove that token 
from it. The Paralysis Token does not prevent Units 
from performing Retaliation Attacks.

### LINKED ISSUE #674: Pit Lords' Sommon ability
There are two pieces of information that are missing in the current version:

1. Inferno rulebook specifically states that Pit Lords' Summon ability can only be used once per combat and that Reinforcement case is free unless you want to pay the cost (page 3, right column, second paragraph)
Screenshot_20240831-151006_Adobe_Acrobat

2. It's clarified by Archon staff on their Discord server that while RAW Pit Lords can Summon a Pack of Demons that are already Recruited but not present on the combat board (the player chose not to draw them for that combat) , RAI it's not intended to work this way. Source: [message link](https://discord.com/channels/740870068178649108/1211591995328299038/1275821683948916808).
Screenshot_20240831-151658_Discord

qwrtln:
Fixes #538
Fixes #674 

en-13
en-33


sections/expansion_content.tex
```diff
@@ -79,10 +79,12 @@ \subsection*{\pagetarget{Events}{Events}}
 
 \subsection*{Summoning}
 Some cards from the Inferno expansion may Summon Units\index{Summoning} during Combat.
-This effect cannot Summon Units from the Neutral Units Deck.
+This effect can be used \textbf{once per Combat}.
+It cannot Summon Units from the Neutral Units Deck.
 Place the summoned Unit adjacent to the summoning Unit.
 Summoned Units Activate in the Round they were summoned if their Initiative is lower or equal to the Initiative of the currently Activated Unit.
 Otherwise, treat them as if they already activated this Combat Round.
+You can also Reinforce the Unit already on the Combat Board with this effect, and may pay no gold.
```
DarthGandalf, line 87: also or alternatively?

Cyber-EX:
@qwrtln  did you forget to commit text about second piece of info, that you cannot summon a Pack of Demons?

qwrtln:
> @qwrtln did you forget to commit text about second piece of info, that you cannot summon a Pack of Demons?

It's not obvious to describe in a generic way. Please suggest your own wording.

Cyber-EX:
@qwrtln For example:
"Pit Lord cannot Summon Reinforced Demons (i.e. Pack) even if Player has Reinforced Demons in their Unit Deck. In this case Demons should be Summoned as Few, but after Combat they return to Player's Unit Deck as Pack."

qwrtln:
> For example: "Pit Lord cannot Summon Reinforced Demons (i.e. Pack) even if Player has Reinforced Demons in their Unit Deck. In this case Demons should be Summoned as Few, but after Combat they return to Player's Unit Deck as Pack."

OK, are you sure this is RAI? You can summon few Demons even if you have a pack already recruited?

Cyber-EX:
I read a few times this line from Inferno Mission Book:
_"After Combat, unless stated otherwise, the summoned units are added to your unit deck."_
I think they forgot to say "if they survive the Combat":
_"After Combat, unless stated otherwise, **if the summoned units survive**, they are added to your unit deck."_

Cyber-EX:
> OK, are you sure this is RAI? You can summon few Demons even if you have a pack already recruited?

@qwrtln  I'm judging basing on this - https://discord.com/channels/740870068178649108/1211591995328299038/1275821683948916808 - but after thinking I agree with you that this needs to be doublechecked with official designer.

Cyber-EX:
> > OK, are you sure this is RAI? You can summon few Demons even if you have a pack already recruited?
> 
> @qwrtln I'm judging basing on this - https://discord.com/channels/740870068178649108/1211591995328299038/1275821683948916808 - but after thinking I agree with you that this needs to be doublechecked with official designer.

Ok, it turned out much more complicated than I thought. Better omit this idea from this PR.

sections/player_turns.tex
```diff
@@ -58,7 +58,7 @@ \subsection*{Movement Actions}
 \subsection*{\pagetarget{Town Actions}{Town Actions}\index{Town Actions}}
 You can perform each of the Town Actions listed below \textbf{once per Round}.
 These Actions can be performed at any point during any player's Turn, except during Combat or when your Town Action would interrupt another Action.
-For example, you cannot draw Spell cards simultaneously with the Spell Book Token.\par
+For example, players cannot draw Spell cards simultaneously with the Spell Book Token.\par
```
DarthGandalf, line 61: I'm not saying that the sentence was good before, but it seems to say that players are drawing the book token. Maybe add word "using"?

Heegu-sama: > I'm not saying that the sentence was good before, but it seems to say that players are drawing the book token. Maybe add word "using"?

Agreed. "...players cannot draw Spell cards simultaneously **using** the Spell Book Token."

Heegu-sama (approved):

qwrtln:
po4a commit

qwrtln:
@DarthGandalf Since many of the index entries became fuzzy, I had to overlap index content with the header in RU, as due to alphabet differences it overspilled to the next page. Fix it at your convenience.

PR #797, opened 2025-04-08
Add caching PDF files

qwrtln:
This is the speed-up I was thinking about:

image

Granted, this is the best-case scenario which won't happen often, but at least there will be no need to wait for all the other languages to build.


.github/workflows/pdf-build.yaml
```diff
@@ -87,8 +67,46 @@ jobs:
...
+            inputs.language == 'en'
+            && hashFiles('main_en.tex', 'metadata.tex', 'sections/*.tex', 'assets/**/*', 'latexmkrc', 'index_style.ist')
+            || hashFiles('main_${{ inputs.language }}.tex', 'metadata.tex', format('sections/translated/{0}/*.tex', inputs.language), 'assets/**/*', format('translations/assets/{0}/**/*', inputs.language), 'latexmkrc', 'index_style.ist')
+            }}
+          restore-keys: |
+            pdf-${{ inputs.build_type }}-${{ inputs.language }}-
+
+      - name: Cache LaTeX packages and build
+        uses: actions/cache@v4
+        if: steps.cache-pdf.outputs.cache-hit != 'true'
```
DarthGandalf, line 88: shouldn't the step below be using the id of this step instead?

qwrtln: This step caches only latexmk artifacts, so it does not replace the build, only speeds it up a little. It shouldn't affect whether the build runs or not.

DarthGandalf (approved):

Heegu-sama (approved):

PR #799, opened 2025-04-14
Upload the patch artifact from failed po4a check

DarthGandalf:


.github/workflows/po4a-check.yaml
```diff
@@ -25,5 +25,14 @@ jobs:
           if [[ `git status --porcelain translations` ]]; then
             echo "There are uncommited changes to translations:"
             git status --porcelain translations
+            git diff > po4a.patch
```
qwrtln, line 28:
```suggestion
            git diff | tee po4a.patch
```

qwrtln, line 28:
```suggestion
            git diff | tee po4a.patch
```

qwrtln (approved):

PR #800, opened 2025-04-16
Add cache invalidation workflow

qwrtln:
Introducing a bot which invalidates cache, shall it ever be broken. Let's hope it never needs to be used. Example:

image

Other changes:
1. Moved inkscape artifacts to a separate cache.
2. Fix key for latexmk build artifacts, adjusted paths.
3. po4a diff is now colored.
4. Adjustment in AI instructions.


.github/workflows/po4a-check.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
           if [[ `git status --porcelain translations` ]]; then
             echo "There are uncommited changes to translations:"
             git status --porcelain translations
-            git diff | tee po4a.patch
+            git diff | tee po4a.patch | git --no-pager diff --color=always
```
DarthGandalf, line 28: Is this meant to be a separate line? I didn't know `git diff` accepts anything from stdin. Also in this case need to replace tee with > to avoid duplicate diff, right?

.github/workflows/pdf-build.yaml
```diff
@@ -83,26 +83,36 @@ jobs:
...
+      - name: Cache Inkscape artifacts
+        if: steps.cache-pdf.outputs.cache-hit != 'true'
+        uses: actions/cache@v4
+        with:
+          path: |
+            ./svg-inkscape
+          key: >-
+            inkscape-${{ hashFiles('assets/glyphs/*') }}
+          restore-keys: |
+            inkscape-
```
DarthGandalf, line 95: This would work as long as svg files don't contain any language-specific text, right?

.github/workflows/invalidate-cache.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: List caches
+        run: |
+          gh api repos/${{ github.repository }}/actions/caches | jq -c '{caches: [.actions_caches[] | select(.key | startswith("${{ env.CACHE_PREFIX }}"))]}' > caches.json
+
+      - name: Add comment
+        run: |
+          if [[ $(jq '.caches | length == 0' caches.json) == "true" ]]; then
+            echo 'No caches found matching prefix `${{ env.CACHE_PREFIX }}` 🤨' >> body.txt
+          else
+            CACHE_COUNT=$(jq '.caches | length' caches.json)
```
DarthGandalf, line 53: First find CACHE_COUNT, then compare it to 0 in if? Now you're counting it twice

DarthGandalf (approved):
Perhaps it would be useful to mention the possibility of adding such comment to the workflow of building the pdf, if building failed?

.github/workflows/po4a-check.yaml
qwrtln: The purpose is to show a colorized diff and redirect an untainted diff to the patch file. It works as intended, but now that I look at it, the `--no-pager` might be redundant. Yes, `git diff` accepts input from stdin, I learned that recently.

.github/workflows/pdf-build.yaml
qwrtln: Correct. AFAIK, they don't.

.github/workflows/invalidate-cache.yaml
qwrtln: True, didn't notice. Those `jq` calls are cheap anyway, so I'm inclined to leave it as is.

.github/workflows/po4a-check.yaml
DarthGandalf: I'd like a proof that it does. In my test which I've just done, it doesn't.

qwrtln:
> Perhaps it would be useful to mention the possibility of adding such comment to the workflow of building the pdf, if building failed?

The workflows can fail for many different reasons, I'm afraid there would be a lot of unnecessary noise. There's no way to tell that it's the faulty cache. I do think, though, that a doc with all the available workflow commands (po4a check, po4a commit, Invalidate cache) would be useful.

.github/workflows/po4a-check.yaml
qwrtln: image

DarthGandalf: Do this instead:

```sh
sed -i -e 's/errors/beers/' main_en.tex
git diff
git diff > patch
git reset --hard
git diff
cat patch | git diff
```

In your case it "works" not because git diff reads stdin, but because it reads the state of files again, just like the git diff before it

qwrtln: Git can read from stdin anyway:
```sh
sed -i -e 's/errors/beers/' main_en.tex
cp main_en.tex main_beer.tex
git reset --hard
cat main_beer.tex | git diff --no-index -- - main_en.tex
```
I'm not arguing which case is which, but it definitely works on github. Are you suggesting any change?

DarthGandalf, line 28:
```suggestion
            git diff > po4a.patch
            git diff --color=always
```

.github/workflows/invalidate-cache.yaml
DarthGandalf, line 53:
```suggestion
          CACHE_COUNT=$(jq '.caches | length' caches.json)
          if [[ ${CACHE_COUNT} == 0 ]]; then
            echo 'No caches found matching prefix `${{ env.CACHE_PREFIX }}` 🤨' >> body.txt
          else
```

Heegu-sama (approved):

PR #807, opened 2025-05-18
Add small clarification and fix link

DarthGandalf:
en-all


Heegu-sama (approved):
I modified the clarification about tiles a bit, otherwise approved 👍

PR #814, opened 2025-06-04
Add trailing whitespace check

qwrtln:


.github/workflows/check-trailing-whitespace.yaml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    name: Check trailing whitespace
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Find trailing whitespace
+        run: |
+          ERRORS=0
+          for f in $(git grep -Il '')
```
DarthGandalf, line 18: Why so complicated?

Couldn't a simple grep -rI do the trick?

PR #821, opened 2025-06-13
Use SHA to determine cache validity in compare_pages.sh

qwrtln:


tools/compare_pages.sh
```diff
@@ -64,22 +64,52 @@ file_mod_time() {
...
+
+  if [[ ! -f "$pdf_file" ]]; then
+    return 1
+  fi
+
+  # Check if poppler is available
+  if command -v pdftotext >/dev/null 2>&1; then
+    # Get 7-character commit SHA from main branch (matching PDF format)
+    local pattern
+    pattern=$(git --no-pager log main -1 --format="%h" 2>/dev/null)
```
DarthGandalf: I can easily have my local version of main branch outdated, then this will claim that the file is always up to date

qwrtln: Maybe that's a pretty niche case, but it's annoying to download every hour. I changed to firstly check the time, and then compare it with remote SHA, would that work?

qwrtln: And/or increase it to 3 hours? 1h is too short.

tools/compare_pages.sh
```diff
@@ -64,22 +64,54 @@ file_mod_time() {
...
+  if [[ $age -le 3600 ]]; then
+    return 0
+  fi
+
+  # File is older than 1 hour, check commit SHA if possible
+  if command -v pdftotext >/dev/null 2>&1; then
+    # Get 7-character commit SHA from remote main branch (matching PDF format)
+    git fetch origin main 2>/dev/null
+    local pattern
+    pattern=$(git --no-pager log origin/main -1 --format="%h" 2>/dev/null)
```
DarthGandalf: And now your assumption is that this remote is called 'origin'. It can be called 'upstream' or anything else really

What is the issue with the time-based approach?

qwrtln: It's just annoying, slow and too frequent, especially for printable.

tools/compare_pages.sh
DarthGandalf: LGTM to just increase it to 6 hours, without touching SHA.

PR #827, opened 2025-07-03
Add all new components to components chapter

DarthGandalf:
en-01
en-02
en-03
en-04


DarthGandalf:
@DerSchlupf all done except links. Because link targets are still in separate PRs, it's hard to test

qwrtln:
> Please clarify.
> 
> I'm already using it for tokens which are overlapping.

Ah, sorry, my impression was that [sub]figures dominate the layout, which would enforce a table-like structure. I was thinking to make the whole page a one big tikzpicture, to make the layout more funky. There's still some underutilized negative space.

qwrtln:
Regarding the minis, this might work for the digital version, but those pixelated ones will be visible in print:

image

The good-looking ones are (from the left) magic elemental, crew mate, cyclops, and archdevil. The other ones should either come from different, higher-res sources, or at the very least could use some upscaling.

DarthGandalf:
Invalidate cache texlive-*

qwrtln:
Invalidate cache pdf-printable-pl-

qwrtln:
Invalidate cache texlive-printable-pl-

PR #832, opened 2025-07-06
[DE] Update all map-locations

DerSchlupf:
WiP: I've started with updating map location chapter. Continue next days...


translations/all_map_locations.tex/de.po
```diff
@@ -578,69 +442,79 @@ msgid ""
...
 "\n"
-"  \\phantom{j}\\textbf{Windmühle}\\\\\n"
-"  \\framedimage[\\linewidth]{\\map_locations/windmill.jpg}\n"
-"  {\\small Achtung Druckfehler: Die Windmühle auf Gebiet \\texttt{\\#N3} hat ein \\textbf{?}. Es muss ``\\textbf{+}1 \\svg{valuablegreater}'' lauten.}\n"
+"  \\phantom{j}\\textbf{Boje}\\phantom{j}\\\\\n"
+"  \\framedimage[\\linewidth]{\\map_locations/buoy.jpg}\\\\\n"
+"  \\expansionlabelnocap{cove}\n"
+"\n"
+"  \\bigskip\n"
+"\n"
```
qwrtln, line 504:
```suggestion
"\n"
" \\vspace*{1.45em}\n"
```

qwrtln:
I'll fix the gate in the original after our PRs are merged.

translations/all_map_locations.tex/de.po
qwrtln: Because of umlauts, this value might need to be lower. Try 1.44 or 1.43.

DerSchlupf:
po4a commit

DerSchlupf:
Because of the thicker symbols the layout doesn't fit. Cause the orange marked paragraph has three lines instead of two. See picture below. Its hard to short it without using less nice wording. Wdyt about removing /medskip (or probably replace it mit /smalkskip) to move the whole text above (green arrow)? Would that fit. What are your experiences?
Screenshot_20250708_174120_com_google_android_apps_docs_PdfViewerActivity

translations/all_map_locations.tex/de.po
```diff
@@ -190,18 +147,18 @@ msgid ""
 msgstr ""
 "\\begin{itemize}[itemsep=0.4em]\n"
```
qwrtln, line 148:
```suggestion
"\\vspace*{-0.8em}\n"
"\\begin{itemize}[itemsep=0.4em]\n"
```
It's not the symbol's fault, as the original also has 3 lines. It's the question mark's explanation which got longer, and has one more line in German. We can trick LaTeX into making the layout tighter.

DerSchlupf:
po4a commit

translations/all_map_locations.tex/de.po
DerSchlupf: I mean in the screenshot in the starting post the german translation of that text is in two lines. But i try the shorter text first. if it doesnt work, i will trick LaTeX.

DerSchlupf:
Okay, i think i'm ready. All final pics in the first post. Do you think here is something to do or is it ready for merge?

PR #836, opened 2025-07-11
Improve economy build

### LINKED ISSUE #487: Colorblind-friendly unit tiers
Выделение_569
изображение

Выделение_570

WDYT?

If yes, I'll make necessary changes to make this work across the whole book

### LINKED ISSUE #779: Economy printable: remove more art
en-10

I'm thinking what to do with non-framed silhouettes

WDYT?

qwrtln:
For the economy build, this change:
  - Removes background from notes and tables
  - Adds letters to unit glyphs
  - Removes backgrounds from section headers
  - Makes yellow glyphs black
  - Changes white text to black
  - And unifies all glyphs under a single svg command

Fixes #487, fixes #779 (?)

Nothing gets fuzzy, no change to the color builds. Select diff:
en-04
en-10
en-12
en-13
en-26
en-29
en-31
en-43
en-47
en-61


metadata.tex
```diff
@@ -229,15 +229,45 @@
 }
 \ExplSyntaxOff
 
+\newcommand{\ifunit}[3]{%
```
DarthGandalf: wouldn't this be simpler with latex3 case?

Or even use the stringswitch function which is defined just above

metadata.tex
```diff
@@ -229,15 +229,45 @@
...
+  \IfStrEq{#1}{bronze}{#2}{%
+  \IfStrEq{#1}{silver}{#2}{%
+  \IfStrEq{#1}{golden}{#2}{%
+  \IfStrEq{#1}{azure}{#2}{#3}%
+}}}}
+
+\newcommand{\ifnoraisepfx}[3]{%
```
DarthGandalf: what's pfx? please add a comment what this is doing / how to use it

metadata.tex
```diff
@@ -229,15 +229,45 @@
...
+    }%
+  }{%
+    \edef\svgpath{\svgs/\detokenize{#2}.svg}%
+  }%
+  \newif\ifnoraise
+  \noraisefalse
+  \ifnoraisepfx{#2}{\noraisetrue}{}%
+  \ifnoraise
+    \includesvg[height=#1px]{\svgpath}%
+  \else
```
DarthGandalf: you used all kinds of "if" here :) I see one of them ends with `}` as part of `iftoggle`, another ends with `\fi`.

This \else is for which if?

In other words, I have no idea what is the control flow here and what this does

metadata.tex
qwrtln: Prefix, yeah, it's redundant, I'll remove it.

metadata.tex
qwrtln: `\iftoggle` doesn't have an explicit else clause, I think? It's for the other one. I'll try adding some comments.

metadata.tex
qwrtln: I couldn't do it with string matching, it would just resolve as false every time. If you have an idea how to do it with latex3, I'm open for suggestions.

metadata.tex
DarthGandalf: I see several statements which may be the \if but none of them is. If it's `\ifnoraise`, I see it several times here, which doesn't match the number of elses.

Ah, the one after \newif is the definition? Is it any better than toggle? just trying to avoid mix of the different styles

qwrtln: I see toggle used here as an external if (by convention, not necessity), the other ones are internal.

PR #838, opened 2025-07-16
Morale Cards in regular games

DerSchlupf:
TO DO:
- [x] Clarfications included (see Dev-comments below the post)
- [x] Morale Cards asset included
- [x] Proofreading (is it done?)
- [x] Discuss the order

en-13
en-14

### Clarifications:
**1)** Source: [https://boardgamegeek.com/thread/3302528/article/44291691#44291691](url)
image 
**2)** Source: [https://discord.com/channels/740870068178649108/1211591995328299038/1394584214296330251](url) image


DerSchlupf:
missing asset:
image
(Source: https://archon-studio.com/files/manuals/homm/homm-battlefield-rulebook_EN.pdf)

DerSchlupf:
What about changing the order: put moral cards directly after moral action and the example at the end? You could remove the shield art and replace it with moral cards introduction and assets. Rest of moral cards on the next page. My main reason is that moral cards and tokens belong directly together.

DerSchlupf:
From discord:
> Regarding morale cards being placed where they are, my thinking was that it's an optional rule from a hardly played expansion. This is why I put it after gameplay example, which IMO should take precedence (it doesn't mention morale anyway, though it could?).
On the page with the shield, there's not enough room for the whole morale cards section, and I wouldn't want to split it.
And even if we were to fit it on that page somehow, good luck translating it to German

I personally still prefer to have the two Morale Systems next to each other, but I can get used to this version, too.

I also have another presentation in mind, but I haven't found a good solution yet, and then the paragraph about the moral tokens would have to be rewritten a bit. That doesn't have to happen right now. Perhaps we can address it in a new PR at some point.

So either you're still discussing the issue of the order, or it's ready for asking Heegu about his opinion and probably then ready for a merge?

PR #839, opened 2025-07-16
Expansion Map Elements

DerSchlupf:
**WiP!!**

To do for the three documents in this PR:

### **map-elements:**
TO-DO:
- [x] asset elemenatl map tiles
- [x] assets location token
- [ ] proofreading (also Links, Index, etc.)
- [ ] missing clarifactions: hero position on elemantel map / discvover between surface & subterranean

en-29
en-30

### **expansion content:**
TO-DO:
- [x] decide which assets and include them (possible ones: creature bank token/backside; stack token; creature banks unit cards; creature bank combat setup)
- [x] general layout
- [ ] proofreading (also Links, Index, etc.)

en-49
en-50

### **all map locations:**
TO-DO:
- [x] fill space next to tavern
- [x] repair storm elemantal art (p.56) and cyclops (p 59.)
- [x] include padoras box variant
- [x] make location pics for whirlpool and creature banks (p.60)
- [x] add the right texts to whirlpool and creature bank introduction
- [ ] proofreading (also Links, Index, etc.)
- [ ] clarifications: whirlpool "losing unit", monoloith MP cost; pandoras box card;

en-61
en-62
en-63


DerSchlupf:
By the way i think these assets should we add: 
image
image

qwrtln:
Added assets, how's that?

en-04
en-05

DerSchlupf:
Some ideas:

1) replace "stronghold map tile" with "surface map tile" cause the difference between surface and subterranean is relevant, not the expansion, where the tile comes from. 

2) should we list the token heading at first? It's kind of a general rule and after that we write something about specific tokens (whirlpools, gate...).

3) We could add a sentence about tokens under the "discover and placing tiles" heading like "Apply further rules, when discover a map tile with location tokens". (and add a link to the token chapter).  But probably it's not necessary.

4) I will push a suggestion about including creature banks and elemental map tiles here as well soon.

DerSchlupf:
--> changed the order, added heading location token, added whirpoll (without asset unfortunatly) and create pagetarget at creature bank table (red marking) 
--> removed firebird art, but there is space for new art next to tavern.

Should i push it? 

image
image
image
image

DerSchlupf:
Also in map elements: 
- changed the order (the general Token Text i would put first, wdyt?)
- renamed stronghold map til in "Surface Map Tile"
- whirlpool text removed (is now in all map location) 
- made subterra gate text shorter, but not removed cause the rules are more tricky.
- added couple of pagetargets and -links between all map location and map elements (and creature bank expansion content)

Push?
image

DerSchlupf:
There is supposed to be one "}" too many in line 751 or what don't i understand. I found no error in the code.  Probably you can repair it.

DerSchlupf:
By the way: The way it's structured now, it might make more sense to swap the two sides, so map tiles first and tokens afterward. Like you've suggested before :D Decide you what you like more.

DarthGandalf:
The order of images of elemental tiles doesn't match the order of their labels. I think we should update the asset to swap the air and water tiles. It's possible to swap labels instead, but that removes consistency with other labeled images throughout the book, which are in 2 columns

Thoughts?

DarthGandalf:
Or even like this:
blah

DerSchlupf:
Suggestion for text changes:

**Discover token:** 

Location token are added by several expansions to the game.  They are usually placed on specific fields of the map, when a hero discovers the associated map tile and overwrite the original location they are placed on. But they also might have an own position in the map layout, depending on the chosen scenario. Location Tokens *cannot* be placed on blocked fields, other location tokens or fields containing locations required to meet any of the scenarios victory condition. Keep this rules in mind when rotating map tiles with location tokens.

**Subterranean gate:**

Subterranean gate token are also determined by the scenarios map layout and they connect both a surface map tile and an adjacent subterranean map tile. After discovering the first of the corresponding map tiles you may choose the position of the subterranean gate token as long as it connects the two indicated map tiles and without covering forbidden fields. The scenario map Layout only shows  **which two tiles are connected**, not the exact fields that must be connected.

**Creature banks**

As an **optional rule** you can decide to play with creature banks. If you do each time a player discovers a far or near map tile they may decide to draw a creature bank token corresponding to the discovered map tile and – contrary to the normal rules – place it on its blocked field.

DerSchlupf:
And regarding subterrain gate in all map location:
Screenshot_20250804_155053_com_android_chrome_ChromeTabbedActivity

That part with "move to the connected location..." makes less sense. It refers to archons wording with "entrance" and "exit" of such a gate what we skipped in the rewrite. 

**Possible other wording:**

Connects a surface and a subterranean Map Tile (Link to subterranean map tiles). Treat both fields of the gate as single field. If the connected Tile isn't discovered, discover it. Otherwise treat this location as empty field.

DerSchlupf:
And last but not least regarding this:
Screenshot_20250804_222303_com_android_chrome_ChromeTabbedActivity~2

I would suggest to add "drawing a token", e. g.:

... you may choose to draw a Creature Bank Token of the corresponding type and place it on one of that's tiles blocked fields.

(or something similar)

Heegu-sama (approved):

Issue #628, opened 2024-07-01
Add expert effect exception for Schools of Magic


captain-yoshi:
Based on this discord [thread](https://discord.com/channels/740870068178649108/1211591995328299038/1257066534933762108), it seems that we can use the Expert Effect on schools of magic abilities when already in play. If true, we should add some text or an amber box for this exception.

![image](https://github.com/Heegu-sama/Homm3BG/assets/32679594/50a1501a-87c8-443c-a434-81cf3472a420)

qwrtln:
Fixed by #842

PR #842, opened 2025-07-22
Mention time tokens in deck-building

DerSchlupf:
Two additions:

### 1. Mention Ongoing Token
TO-DO:
- [x] Include Token Asset
- [ ] Discuss, which kind of expansion mark should be there
- [ ] Discuss, if it is enough info for that token.

Honestly speaking i don't think the official rule about time tokens is really helpful. Proabably better just to mention to show what this component is for but let players handle it for themselves.

### 2. Update Permanent Card (would solve https://github.com/Heegu-sama/Homm3BG/issues/628)
TO-DO:
- [x] Clarification about basic and expert effect added:
image

- [ ] Discuss, if clarification about map/activation-effect is needed
image

CURRENT LAYOUT

en-18
en-22
en-46
en-48


qwrtln:
en-01

DerSchlupf:
> Time token was in several expansions, iirc

We should fundamentally clarify how such content is handled. In my opinion we have four possibilitys:
A) **We dont mark them** as expansion conent. That's that, what we still do with "permanent Cards".
B) **We mark them with all relevant expansions**. That would be much text since much content comes from many (or all) expansions. I wouldn recommend that.
C) **We mark them with one expansion** and and keep quiet about the others. That could be confusing, cause people believe they need this one specific expansion.
D) **We mark them with a new bar called "Several expansion" or something similar**. This bar should be explained in the introduction as well.

I prefer version A) or D). In both cases, the chapter "Expansion content" will give an overview about all expansion content and also such content, which is marked like A or D.

PR #844, opened 2025-07-26
New Section "Detailed Card Effects" for specific token rules

DerSchlupf:
### WiP!!

Here a suggestion to include all the rules about specific cards and their own components into the rewrite. The idea is to create a new section. Please say honestly, if such a new section isn't needed. In this case, we can copy & paste the written texts and move them whereever they should be.

TO-DO:
- [x] Discuss this concept of a new section
- [x] Discuss the Section header 
- [x] Repair Layout
- [x] Proofreading
- [ ] Are there clarifications needed? If yes, Add missing clarifications
- [x] Add clone example
- [x] Make clone example image
- [ ] Index printable of that keywords?
- [x] Add Link/ Note in Deckbuilding Section (Spells)
- [x] Add Link/ Note in Unit section
- [x] Remove Summon demons from unit section

Links in Deckbuilding and Unit chapter:
en-21

Units chapter:
en-32
en-33
en-34

Expansion content chapter:
en-62
en-62-63

New chapter:
en-77
en-78
en-79
en-80


DarthGandalf:
Not sure about the new image: for me it reads like you're allowed to cast Haste on the original unit, but not on the clone.

DarthGandalf:
With force field and land mine, and manticore, probably the sentence in Expansion Content that pandora box is the only gameplay element from SG2 is not true anymore?

DarthGandalf:
I don't think attack and weakness should mention the unit. These 2 tokens are mentioned in SG2 book too, not only in stronghold/cove. So probably something unknown in SG2 has a way to add these tokens without being Sorceress and Ogre.

Then this becomes not "Unit Ability Keywords", but just a bunch of random tokens. Perhaps name the chapter "Additional Tokens"? And move Summon Elemental to be near Summon Demon in Units chapter

WDYT?

DarthGandalf:
force field: if you think the interaction between quicksand and force field is clear, ok

weakness: we shouldn't invent randomness until seeing the text on the card

I think both summons (demon and elemental) should be on the same page, because they are so different, so that the reader doesn't stumble upon the wrong one, and uses the rule from wrong one

DarthGandalf:
> Also here i wouldn‘t do that, cause summen elementals is a complex spell Token

I wouldn't put everything which happens to use tokens to the same chapter, but to what this thing is about.

DerSchlupf:
> Here's draft. The Summoning is in Units

I think it makes more sense to have all the rules for specific cards in one place. Why should players look for the rules for a spell under "Units"? Or why not the rules for "cloning units" as well? That adds a unit to the combat board, too.

I thought summon demons was there cause it is a unit ability.

DarthGandalf:
Updated images in first post, PTAL

DerSchlupf:
I like your last changes! Good idea, to visulize the different unit cards in the  Unit section. What about adding a header, but an old one? e. g. "Different Unit Card Types" or somethin similar?

DerSchlupf:
Hmm, i looks a bit weird to have the card picture at first without any text before. but i havent a good idea. We could also include the front side next to the back side. maybe it looks better. but we could also keep it as it is. it's not so important. 

But i have to other questions:

### 1) What about Links in Spell und Unit ability chapter?
image

### 2) What is the benefit of this table?
I know, the original has it, but in my opionin it is not really helpful. But if we keep it (what i guess we do) shouldnt we include all expansion faction units into this table? maybe switching x- and y-axis and having one faction per row and a longer table. 
image

DarthGandalf (approved):

PR #846, opened 2025-08-08
Update ua.po

MarkGrasteinn:
Updated translation to "Introduction" block of Ukrainian language according to August 5, 2025.

### DIFF
--- a/translations/introduction.tex/ua.po
+++ b/translations/introduction.tex/ua.po
@@ -42,19 +42,21 @@ msgid ""
 "their Heroes, battle for supremacy. Choose your Faction and Hero and banish "
 "your unruly enemies from these lands!"
 msgstr ""
-"На континенті Антагаріч йде війна: Фракції на чолі зі своїми Героями "
+"На континенті Антагаріч іде війна: Фракції на чолі зі своїми Героями "
 "борються за верховенство. Оберіть свою Фракцію та Героя і виженіть своїх "
 "непокірних ворогів з цих земель!"
 
 #. type: multicols*
 #: sections/introduction.tex:11
 #, fuzzy, no-wrap
-#| msgid "\\pagetarget{Heroes of Might and Magic III}{\\textbf{Heroes of Might and Magic III: The Board Game}} is a tactical strategy RPG board game for 1--3 players using the core box set, or more when using expansions."
 msgid ""
 "{2}\n"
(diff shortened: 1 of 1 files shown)


translations/introduction.tex/ua.po
```diff
@@ -84,6 +93,10 @@ msgid ""
 "\\end{expansion}\n"
 "\n"
 msgstr ""
+"\\begin{expansion}{tower}\n"
```
DarthGandalf, line 96: You don't need to touch anything in `\phantom{}` as it'll have very little effect. In fact, this particular part of Intro doesn't need to be translated at all, as the only translateable part here is headers of expansions, but translation for headers in the expansion blocks are defined in https://github.com/Heegu-sama/Homm3BG/blob/main/main_ua.tex#L46

See e.g. https://github.com/Heegu-sama/Homm3BG/blob/main/main_ru.tex#L48

translations/introduction.tex/ua.po
```diff
@@ -84,6 +93,10 @@ msgid ""
...
+"  \\phantom{Доповнення "Башта".}\n"
```
qwrtln, line 97:
```suggestion
"  \\phantom{Доповнення ``Башта''.}\n"
```
Do this for other expansions too.

qwrtln: It's not important here, but in other cases those apostrophes will be.

DarthGandalf: That would produce wrong quotation marks, which are used in English.

From what I can see in https://en.wikipedia.org/wiki/Quotation_mark#Belarusian,_Russian,_and_Ukrainian « » is used in Ukrainian. In Russian I'm using `<< >>` here which LaTeX transforms into « » but using Unicode characters themselves is fine too.

translations/introduction.tex/ua.po
```diff
@@ -174,34 +219,14 @@ msgid ""
...
 "\n"
 msgstr ""
+"\\begin{expansion}{stretchgoals2}\n"
+"  \\phantom{Доповнення "Цілі кампанії 2".}\n"
+"\\end{expansion}\n"
+"\n"
 
 #. type: multicols*
 #: sections/introduction.tex:76
 #, fuzzy, no-wrap
```
qwrtln, line 229:
```suggestion
#, no-wrap
```
You have to remove `fuzzy` for the translation to be included.

PR #852, opened 2025-08-25
[DE] Translation Update Map Elements & Creature Banks

DerSchlupf:
some translations (eg creature bank tables) still missing.  i need layout to fix.

MAP ELEMENTS
image

EXP CONTENT
image

MAP LOCATIONS
image
image


DerSchlupf:
Translation of these party are done. The final screenshots are in the first post.

But no idea how to short the text that it is still clear AND fits in the cell in this table. i reduced the scriptsize to solve the problem. Do you find this annoying? Do you have another idea? For me it would be okay and then it is ready to merge, I think.

image

PR #853, opened 2025-08-25
[WIP] Add bleed support for printable mode

captain-yoshi:
Add support for bleed in printable mode. Default bleed size is set to 3mm (standard). In this mode, the actual page size will grow from 210 x 297 mm to 216 x 303 mm. Setting the bleed to zero is equivalent to the printable mode we had since the begining.

<p align="center">

</p>

The only hiccup is the bottom layout :
(digital)
image
(bleed + printable)
image

The bleed lowers the bottom layout and it does not match the original. But the digital and printable bottom are different to begin with (smaller in width, that is why the printable image is distorted). 

Remaining things to do:
- [ ] use the original bottom layout ? and shift for odd/even pages and expand around 35px (for 300 DPI) vertically so that it is more akin to the original. This would be *harcoded* for 3mm bleed at 300 DPI.
- [ ] Recheck the *-p|--printable* console option, did not work for me...
- [x] The cover images and other backgrounds relying on *\paperwidth* are a little bit translated by the bleed, but I don't think it's worth the effort. It's hard to see the difference.

### DIFF
--- a/metadata.tex
+++ b/metadata.tex
@@ -1,14 +1,50 @@
-% Set page size and margins
-\usepackage[
-  a4paper,
-  top=2cm,
-  bottom=2.3cm,
-  left=2cm,
-  right=2cm,
-  marginparwidth=1.75cm,
-  footskip=1.35cm,
-]{geometry}
+\usepackage{etoolbox}
+\newtoggle{printable}
+\newtoggle{noartbackground}
+\newtoggle{githubbuild}
 
+
+% Computes trimmed size and bleed margins (for print) or plain A4 (for digital)
+\newlength\trimW \setlength\trimW{210mm}
+\newlength\trimH \setlength\trimH{297mm}
+\newlength\bleedTop    \setlength\bleedTop{3mm}
+\newlength\bleedBottom \setlength\bleedBottom{3mm}
+\newlength\bleedLeft   \setlength\bleedLeft{3mm}
+\newlength\bleedRight  \setlength\bleedRight{3mm}
+
+\edef\GEOpw{\the\dimexpr\trimW+\bleedLeft+\bleedRight\relax}
+\edef\GEOph{\the\dimexpr\trimH+\bleedTop+\bleedBottom\relax}
+\edef\GEOlw{\the\trimW}
+\edef\GEOlh{\the\trimH}
+\edef\GEOhx{\the\bleedLeft}
+\edef\GEOvy{\the\bleedTop}
+
+\iftoggle{printable}{%
+  % Printable build with bleed
+  \usepackage[
+    layoutwidth=\GEOlw, layoutheight=\GEOlh,
+    layouthoffset=\bleedLeft, layoutvoffset=\bleedTop,
+    paperwidth=\GEOpw, paperheight=\GEOph,
+    twoside,
+    top=2cm,
+    bottom=2.3cm,
+    left=2.5cm,
+    right=1.5cm,
+    marginparwidth=1.75cm,
+    footskip=1.35cm,
+  ]{geometry}%
+}{%
+  % Digital build (no bleed)
+  \usepackage[
+    a4paper,
+    top=2cm,
+    bottom=2.3cm,
+    left=2cm,
+    right=2cm,
+    marginparwidth=1.75cm,
+    footskip=1.35cm
+  ]{geometry}%
+}
 % Useful packages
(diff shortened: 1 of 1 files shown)


metadata.tex
```diff
@@ -1,22 +1,57 @@
...
+
+% Computes trimmed size and bleed margins (for print) or plain A4 (for digital)
+\newlength\trimW \setlength\trimW{210mm}
+\newlength\trimH \setlength\trimH{297mm}
+\newlength\bleedTop    \setlength\bleedTop{3mm}
+\newlength\bleedBottom \setlength\bleedBottom{3mm}
+\newlength\bleedLeft   \setlength\bleedLeft{3mm}
+\newlength\bleedRight  \setlength\bleedRight{3mm}
+
+\edef\GEOpw{\the\dimexpr\trimW+\bleedLeft+\bleedRight\relax}
```
DarthGandalf, line 15: This makes the page bigger than A4, so that it after cutting the result is A4?

My issue with A4 pages was that they don't fit in the game box (I don't have the big box, it wouldn't be an issue there). To print my copy of the book, my workaround was to downsize all pages after pdf was produced, to be able to cut excess, making the paper a bit less than A4, then it fits.

captain-yoshi: Yes, they print it on SRA4 (225 x 320 mm) or similar format and then get trimmed down to the finished size, so A4 format. It would be too difficult for any printer to try and print directly on a A4 paper (you would sometimes see white borders depending on the alignment).

That's a valid point for not fitting in the game box. Yeah there's 7 to 10 mm to trim for it to fit in the game box. But if you remove that vertical space, it hinders the original layout. You would had to reduce some images size or  paragraph spacing. Yeah, downsizing is easier for keeping the original aspect of the rulebook intact.

So using ghostscript, you could still keep bleed and remove 10mm and it would fit in the gamebox. Probably what you did. But it would cost you more because its not a standard size, but would be really close to the original. 
```bash
# -g uses 1/10 pt: 206 mm → 5839.36 pt → 5839, 293 mm → 830.552 pt → 8306
gs -o output.pdf -sDEVICE=pdfwrite \
   -dFIXEDMEDIA -dPDFFitPage \
   -g5839x8306 \
   main_en.pdf
```
You could also print it without bleed on an A4 paper with downscaling and trim the rulebook manually.

qwrtln:
> I'm confused, what is the intended definitive paper size for this rewritten rulebook ? I though it to be A4. If you want it at the non-standard size, you gotta post-process and/or do some scaling to achieve the same thing as the digital version.
> 
> I guess you had in mind that the printed version should be like the official booklet. It's a valid point, and I'm not against it, it's just much more work (unless you scale it to retain the same effect). I really don't mind if you don't want this to be added to the project. Regardless, you still need bleed if you want to print it professionally.

Yes, it's much more work, and that's exactly the point. We don't want to achieve this with scaling or other trickery. If we want to add bleed, I'd expect it to be added the way a professional pressman would do it with Adobe or otherwise.

> I will recheck, but everything relying upon, e.g. _\textwidth_ and _\linewidth_ should follow the trim (original paper size) section. It will have the same proportions as the original. Only things like background images and the bottom or anything relative to the paper, e.g. _\paperwidth_ will bleed out.
> 
> For these images, they could be repositioned so they do not appear in the bleed, but some images, e.g. the bottom layout don't have enough information to position correctly, we must use other techniques to add missing information, and if the bleed size changes, the missing information must be added dynamically (and it also depends on the DPI) used. I would propose to add a standard bleed in metric and in imperial for these cases.

No repositioning is necessary, as we had none of that with professional printing we did last year.

image
image

I can't find better pictures now, but you can see that the bottom margin is visibly trimmed (so it can be bled out), while the left and right one are still quite wide. The size matches the original 280x198 mm, and the booklet fits in every game box.

The point is that the bleed lines should not touch the actual bleed, and the sides should be padded with previous/next pages, like here:

image

> Are you referring to the official booklet ? In this PR, I keep the same margin, it is relative to the trim part not the bleed part, so the text and images like the section heading will be like the original digital version. The printed version is just offset for even/odd pages.

I'm not sure what you mean by that, maybe I'm confused by the terminology here. My point is that having the printable version page size match exactly the digital one is not the goal.

BTW, there's a separate issue of the cover. If you want to make it a perfect bound booklet as we did (rather than a folded leaflet like the original), it's another consideration.

PR #867, opened 2025-09-09
WiP: Add Tournament rules?

DerSchlupf:
I've written the rules for tournament mode, for now as a new section. I've included the rules for splitted artifact and spell deck in this section, even if we move them later in other chapters. I've chosen a slightly different structure than the original tournament book, but i#M still not sure, if it makes sense how it is :D The two tables are just a try to visualize. I like them more than a list, but a list is also included in the PR (marked as %-Notes).
Furthermore, i havent play tournament mode, yet. So i dont know, if it is clear written or if we need some more clarifications. At the end, this pages are also still highly work in progress. I found two clarifications about splitted decks, see below the current layout. 

CURRENT LAYOUT:
image
image

CLARIFICATIONS:

1. For effects outside the map, map parts are still crucial. (Source: https://discord.com/channels/740870068178649108/897475192677355560/1201802322624122930)
image

2. But gaining cards outside the map depends only on the MAIN Hero's position (Source: https://discord.com/channels/740870068178649108/897475192677355560/1201813535857057822)
image


DerSchlupf:
Some ideas not to forget:
- Add link in Game Mode section
- Remove redwood observatory location and move all the text in all map location. we have enough place on page 65 to restructure that chapter that both effects of that location are listet there.
- instead of that: mention obersavorys others effect with a link to all map location 
- add link to moral action

DerSchlupf:
Sorry guys, i've created a big chaos here seems so. I will delete this branch complety and create a new one.

PR #880, opened 2025-10-17
some changes in new headers

DerSchlupf:
I've included these changes: 
image

Here the results:

ToC:
image

Player Turns:
image

Deckbuilding:
image
image

Map elements:
image

Units:
image

AI Rules:
image

Expansion Conent:
image
image
image

### DIFF
--- a/translations/map_elements.tex/de.po
+++ b/translations/map_elements.tex/de.po
@@ -83,7 +83,7 @@ msgid ""
 msgstr ""
 "\\vspace*{\\fill}\n"
 "\\begin{center}\n"
-"  \\transparent{0.2}\\includegraphics[width=0.5\\linewidth]{\\art/earth_magic.png}\n"
+"  \\transparent{0.2}\\includegraphics[width=0.4\\linewidth]{\\art/earth_magic.png}\n"
 "\\end{center}\n"
 "\n"
 


DarthGandalf:
en-03
en-13
en-18
en-22
en-23
en-26
en-28
en-34
en-44
en-51
en-52
en-56

sections/deckbuilding.tex
```diff
@@ -46,6 +48,26 @@ \subsection*{General Card Rules}
 \end{multicols}
 \clearpage
 
+\begin{multicols*}{2}
+    \subsection*{Searching Example}
```
DarthGandalf: I think this should still be in toc

sections/map_elements.tex
```diff
@@ -16,7 +16,7 @@
   \transparent{0.2}\includegraphics[width=0.6\linewidth]{\art/earth_magic.png}
 \end{center}
 
-\subsection*{Map Tile Anatomy}\index{Map Tile}
+\subheader{Map Tile Anatomy}\index{Map Tile}
```
DarthGandalf: Considering this is on the same page as Map Elements chapter itself, this seems useless to me

sections/expansion_content.tex
```diff
@@ -378,7 +378,7 @@ \subsection*{More of the same}
 
 \pagebreak
 \begin{expansion}{battlefield}
-  \pagetarget{Battlefield Combat}{\subheader{Battlefield in regular games}}
+  \pagetarget{Battlefield Combat}{\subheader{Combat on the Battlefield Board}}
```
DarthGandalf, line 381: Maybe for now add a note that this doesn't apply to skirmish and adventure?

DarthGandalf (commented):
I agree with the rest

sections/ai_rules.tex
```diff
@@ -5,7 +5,7 @@
 
 \begin{multicols}{2}
 
-\subsection*{AI Hero Turn}
+\subheader{AI Hero Turn}
```
qwrtln, line 6: This text is strangely glued to the header now. Why is that happening? The gap between the text and the header should be equal to the one between the header and the top edge of the page.

DarthGandalf: I was just looking at that. Fixed.

qwrtln:
Maybe those `\iftoggle{printable}{\vspace{-\baselineskip}}{}` should be removed now? They were added with 1.2, because changed geometry messed up those spaces for printable build.

sections/ai_rules.tex
qwrtln: BTW, this is the only subheader I'm not a fan of. It doesn't look right that you have a section header, and then a subheader right below it competing for attention. But I don't have a better idea, so we can leave it at that for now.

sections/expansion_content.tex
qwrtln: @DerSchlupf Was working on battlefield rules, so this shouldn't apply once those changes are added?

DarthGandalf: I don't think he'll be able to finish that by 2.0 though, as he's away for a few weeks.

I'd really like to get 2.0 out by the time the expansions are delivered, with 2.1 coming soon after with new card scans and potentially some more clarifications once people start playing the new expansions and asking questions.

2.1 or 2.2 would probably be a good target for the battlefield rules, so in 2.0 such note would be useful.

sections/ai_rules.tex
DarthGandalf: What about this?

with 2 subheaders it doesn't conflict with main header anymore as much

en-44

qwrtln: Better yeah. Let's hope it doesn't get too tight for translations, as many have struggled with this.

qwrtln (approved):

sections/expansion_content.tex
DerSchlupf: Hi, from today on I will slowly be able to come back to the project. Sry for being quiet the last weeks. 

I've already written the battlefield rules but haven't pushed yet. Today I have the time to bring them in a pushable state and open a new PR. Maybe we have enough time to include it in 2.0.

Thanx for your work the last weeks. Looks fine!

PR #881, opened 2025-10-23
Add link between location tokens and their discovery

DarthGandalf:
en-30
en-69


qwrtln:
I'm not sure if it should work both ways like that. I think the 69-page one should link to map locations, sure, but back?

DarthGandalf:
I think both links are useful:

* what do you do with location after placing it?
* how can I find this location anywhere? No tile has it

I'm not very happy with the text of the links there though.

DarthGandalf:
po4a commit

Heegu-sama (approved):

PR #886, opened 2025-11-09
Make it easier to compare moving pages

DarthGandalf:


DarthGandalf:
An idea: rename the arg from --mapping to --moved?

tools/compare_pages.sh
```diff
@@ -16,6 +16,7 @@ help() {
     Optional Arguments:
       -p, --printable               Compares your build against 'printable' build.
       -s, --single-page             Combines all compared pages into a single image.
+      -m, --mapping A:B             Page A on the left will be with page B on the right.
```
qwrtln: Please add an example, as to me it's not evident how to use it. What if I want to create 5 images, and all of them offset by 1 page?

DarthGandalf: Done

DarthGandalf: For multiple offset pages the syntax is not ideal

tools/compare_pages.sh
```diff
@@ -27,6 +27,13 @@ help() {
           - Then because there is the '--single-page' parameter, it combines them to a single file 'en-all.png'.
           - It will use 'printable_en.pdf' from the repository as baseline because '--printable' was specified.
             It would use 'main_en.pdf' if this parameter was omitted.
+
+      ./tools/compare_pages.sh -l en -r 2-5 --moved 3:8 --moved 4:3
+          This will produce the following 4 images:
+          - en-02.png with page 2 on the left and page 2 on the right,
+          - en-03.png with page 3 on the left and page 8 on the right,
+          - en-04.png with page 4 on the left and page 3 on the right,
+          - en-05.png with page 5 on the left and page 5 on the right.
```
qwrtln: Thanks, I see it now. Wouldn't it be simpler just to have an offset parameter? I guess usually you want to move all pages in either +1 or -1 direction? For X pages there would be no need to specify X mappings. Or even leave mapping if you like, but introduce an additional -o|--offset parameter?

DarthGandalf: This was needed for the headers PR, where some example randomly moved, and I did a single command for all the necessary changed pages

I think usually I need to move pages to some specific location, rather than by the specific offset which user would need to calculate

I've removed this new parameter, and updated syntax of -r, PTAL

tools/compare_pages.sh
```diff
@@ -26,6 +26,14 @@ help() {
           - Then because there is the '--single-page' parameter, it combines them to a single file 'en-all.png'.
           - It will use 'printable_en.pdf' from the repository as baseline because '--printable' was specified.
             It would use 'main_en.pdf' if this parameter was omitted.
+
+      ./tools/compare_pages.sh -l en -r 2,5:7,8-9:6
```
qwrtln, line 31: Consider mentioning this colon syntax in the -r argument explanation.

qwrtln (approved):

PR #888, opened 2025-11-09
Require approval for po4a workflow

qwrtln:


.github/workflows/po4a-comment-commit.yaml
```diff
@@ -61,9 +82,16 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
 
+      - name: Comment about missing approvals
+        if: ${{ failure() && steps.check_approvals.conclusion == 'failure' }}
```
DarthGandalf: Do this directly in that step? Why do you need separate step which checks that one?

qwrtln: It can be exactly in the step, true. I just mirrored what was already there for permissions, where this is the only viable approach.

PR #903, opened 2025-11-20
[FR] All map locations

Legrecofr:
fr-01
fr-02
fr-03
fr-04
fr-05
fr-06
fr-07
fr-08
fr-09
fr-10
fr-11
fr-12
fr-13


qwrtln:
Images require realigning. If you look closely, you can see some of them are not at the same level.

image

PR #913, opened 2025-11-23
Remove leftovers from map tiles image

qwrtln:
en-26


sections/map_elements.tex
```diff
@@ -26,10 +26,19 @@ \subsection*{Map Tile Anatomy}\index{Map Tile}
 }
 \columnbreak
 
-\includegraphics[width=\linewidth]{\images/maptiles.png}
 \begin{center}
+  \begin{footnotesize}
```
DarthGandalf, line 30: footnotesize is now done inside encircle already (in the bf branch...)

qwrtln: Yes, this here is mostly for consistent look.

PR #924, opened 2025-11-27
[DE] Update All Map locations

DerSchlupf:
CUrrent Layout:
image
image
image
image
image
image
image


DerSchlupf:
Hi, all fuzzy translations are done and the first post is up to date. But: The first two pages dont fit but without failing of the "check-pages-check", why? 
I cant short the text more. We can reduce the space between the lines. I dont know why there is so much space in german translation between them. Its quite less in the original. 
image

Can someone help to fit the pages?

qwrtln:
po4a commit

PR #943, opened 2025-12-11
Incorporate devs' clarifications in battlefield rules

qwrtln:
en-56
en-57
en-58
en-59
en-60
en-61
en-62
en-63

### DIFF
--- a/translations/battlefield.tex/cs.po
+++ b/translations/battlefield.tex/cs.po
@@ -90,9 +90,6 @@ msgstr ""
 #: sections/battlefield.tex:30 sections/title_with_toc.tex:37
 #: sections/title_with_toc.tex:42
 #, fuzzy, no-wrap
-#| msgid ""
-#| "\\columnbreak\n"
-#| "\n"
 msgid ""
 "  \\columnbreak\n"
 "\n"
--- a/translations/title_with_toc.tex/cs.po
+++ b/translations/title_with_toc.tex/cs.po
@@ -20,9 +20,6 @@ msgstr ""
 #: sections/battlefield.tex:30 sections/title_with_toc.tex:37
 #: sections/title_with_toc.tex:42
 #, fuzzy, no-wrap
-#| msgid ""
-#| "\\columnbreak\n"
-#| "\n"
 msgid ""
 "  \\columnbreak\n"
 "\n"


sections/battlefield.tex
```diff
@@ -19,29 +19,26 @@
...
-  Victory in Adventure and Skirmish is achieved by defeating all enemy units in the final Combat.
-
   \subsection*{Rule Ambiguity}
   \textbf{Note:} Both game modes were designed for the \textbf{Core Game only}.
-  As a result, the official rules do not address many interactions, particularly those involving expansion content.
-  If you wish to include expansion cards, you will need to establish house rules for unresolved interactions.
-
-  \smallskip
-  Such unanswered questions include:
+  As a result, the official rules do not address many interactions, particularly those involving expansion content. The game developers have \pagetarget{battlefield card clarifications}{clarified the following interactions}:
```
DarthGandalf:
```suggestion
  As a result, the official rules do not address many interactions, particularly those involving expansion content.
  The game developers have \pagetarget{battlefield card clarifications}{clarified the following interactions}:
```

sections/battlefield.tex
```diff
@@ -19,29 +19,26 @@
...
   \begin{itemize}
-    \item Which cards should be removed from the game apart from Diplomacy and Logistics?
-    \item Which cards should be removed in Skirmish Mode because they provide no Combat benefit?
-    \item How should cards with Permanent effects be handled in Adventure Mode?
-    \item How should cards with a ``Remove'' effect (e.~g.~Expert Learning) be handled in Adventure Mode?
-    \item How should cards with both Combat and non-Combat effects (e.~g.~Cape of Velocity) be handled in Adventure Mode?
+    \item The official rules remove only Diplomacy and Logistics ability cards, but common sense applies: the developers also recommend removing Ballistics and Pathfinding, as none provide benefit in Battlefield modes.
+    \item In Skirmish Mode, when a card with no Combat value is drawn, both players may mutually agree to discard it and draw a replacement.
+    \item All cards in the Adventure phase are single-use.
+      Consequently, resource-generating Permanent \svg[7]{permanent} cards (e.g.,~Inexhaustible Cart of Ore) are limited to their bottom \svg{instant} effect only.
```
DarthGandalf, line 39: That's not how I read his answer.

For this specific card, the player is free to choose either 1 stone, or 3 stones, but has to remove the card either way.

For some other card, the choice can be more meaningful

sections/battlefield.tex
DarthGandalf: Let's put pagetarget on the subsection instead. Then the scroll will happen to the header and not to the middle of the text

sections/battlefield.tex
```diff
@@ -19,29 +19,26 @@
...
+      However, other Permanent \svg[7]{permanent} cards (Schools of Magic and War Machines) can be used in the final Combat phase only.
+    \item Cards with ``Remove'' effects (e.g.,~Expert Scholar) may be played during the Adventure phase, applying their effect to your Deck.
+    \item When a card has both Combat and non-Combat effects (e.g.,~Cape of Velocity), you decide whether to use it in the Adventure phase or reserve it for the final Combat.
```
DarthGandalf, line 42: This is one possible interpretation of what Alex said.

Another one is that after you used the card in adventure phase and removed it, it doesn't go back to deck to be used again in combat.

DarthGandalf: Nevermind. Alex confirmed this interpretation

DerSchlupf:
### Regarding 4

My initial reason for cards with remove effect was that balancing thing. See screenshot below again from my reference book.
Screenshot_20251212_170702_com_hihonor_hnoffice_PDFPhoneActivity1

What you've written in that point hasn't to do with the "remove effect" but instead with requiring the keywords "hand" , "discard" pile in a printed effect. Maybe scholar is the only card with that problem, I don't know. Maybe it is confusing to mention that remove effect. What about "if a cards effect works with your hand or discard pile of your deck, you may resolve this effect with your deck instead"

sections/battlefield.tex
qwrtln: Since those aren't strict rules but recommendations, I decided to simplify this:

image

Allowing to resolve "at the beginning of each resources round" as a single-use adventure effect seems like a different class of problems, because I'm not even sure what would that mean? Treat it as an instant effect only once? @DerSchlupf what do you think?

sections/battlefield.tex
qwrtln: okay

sections/battlefield.tex
DarthGandalf: Yeah, for me "a good deal" sounds like a strategy advice, that's why. Not sure what other cases of permanent cards would look like with this

DerSchlupf:
> I'm not sure what you're referring to here. Use the expert effect of [learning](https://en.homm3bg.wiki/abilities/learning)? Why should this be a special case? What makes it so powerful?

In regular games you can play basic learning multiple times, if you draw that card multiple times. So the player can decide to play basic effect to keep that card in his deck, draws it again and plays it a second time, a third time, whatever. So he may generate more XP. If the player plays expert learning, he gets 2 XP in one go, but has to remove that card and looses the ability to play it again. it is similar like an additional cost of that effect. In adventure mode, he may play learning only once, basic or expert, it doesnt matter. So you gain 2 XP, but without paying that additional cost, im mean the cost of doing without the card in the future.

Or take this card. It feels like: Choose one: The worse option or the better option but pay the same^^
image

But You are right because of the crowns. That are a bit more costs. But only for abilities. And its not really an restriction cause you have really only a few set aside cards, at least in that plays, that i have played. And maybe there are some artifacts with the same logic 

But I quote myself :D
> We can leave that point out. Maybe it is not a big thing, which players come up with questions.

maybe it isnt really important.

DerSchlupf:
> Inexhaustible cart of ore is covered in another bullet point as a permanent card - this is allowed, even recommended.

Yeah my list is from an older date. It was without designer comments and only what i've read in the rules.

sections/battlefield.tex
qwrtln: Nothing content here are actual rules, it's all just advice.

DarthGandalf: "The game developers have clarified the following interactions" sounds like rules, not just advice

sections/battlefield.tex
```diff
@@ -19,29 +19,27 @@
...
-
-  \subsection*{Rule Ambiguity}
+  \pagetarget{battlefield card clarifications}{\subsection*{Rule Ambiguity}}
   \textbf{Note:} Both game modes were designed for the \textbf{Core Game only}.
   As a result, the official rules do not address many interactions, particularly those involving expansion content.
-  If you wish to include expansion cards, you will need to establish house rules for unresolved interactions.
-
-  \smallskip
-  Such unanswered questions include:
+  The game developers have clarified the following interactions:
```
DarthGandalf:
```suggestion
  While these are not official rules, the game developers gave several advises on how to deal with this:
```

PR #955, opened 2025-12-16
[HE] update round_structure

davidpodless:
he-14


davidpodless:
No, But I have no idea what caused it. the struct of the msgstr is the same as the msgid

maybe something in the \\begin{expansion}{fortress} is broken? but if that was the case I assume we would have seen in in the introduction

PR #960, opened 2025-12-17
Add wiki links

DarthGandalf:
For printable version there is no difference. User can still click the link if finds one, but it's not visible.

I need to go through whole book to see whether anything else should be linkified.

en-04
en-51
en-56
en-57
en-81
en-83
en-84
en-87
en-88


qwrtln:
Invalidate cache po4a-

DerSchlupf:
AFter someone confirms this: (https://discord.com/channels/1257428163105919098/1257428576891043930/1460317505011581019) , 
maybe we could update the creature bank table in this PR? I guess it is only switching places of Griffin Observatory with crypt. And rewording the headers: instead of part1/2 we should say Far/ Near Tile Tokens. 

I would think a new PR only for this inst necessary?

PR #968, opened 2025-12-22
[HE] update setup

davidpodless:
regarding naval battles expansion,
"Group Stack Tokens, Far- and Near
Creature Bank Tokens into separate
piles and shuffle them"
far- and near are in the same pile so 2 in total, or in 2 different piles so 3 in total?

he-01
he-02
he-03


davidpodless:
he-10
he-11
he-12

DarthGandalf:
in the picture, the circled numbers on events and astrologers are reversed

davidpodless:
Good catch, tnx!

en-10
en-11
en-12

PR #969, opened 2025-12-23
He/fix fuzzies

davidpodless:
nothing fancy, just 2 fuzzies that was added after we already did those files


translations/language_metadata.tex/he.po
```diff
@@ -77,10 +77,8 @@ msgstr "הרחבת \\textenglish{Battlefield}"
 
 #. type: newcommand{#4}
 #: sections/language_metadata.tex:20
-#, fuzzy
-#| msgid "Naval Battles Expansion"
 msgid "Naval Battles \\mbox{Expansion}"
-msgstr "הרחבת \\textenglish{Naval Battles}"
+msgstr "\\mbox{הרחבת} \\textenglish{Naval Battles}"
```
DarthGandalf, line 81: mbox just makes sure hyphenation won't happen. Here it probably doesn't matter

DarthGandalf (approved):

translations/language_metadata.tex/he.po
davidpodless: it works either way, so 🤷🏻‍♂️

PR #972, opened 2025-12-24
[HE] update scenario_end

davidpodless:
en-50


DarthGandalf (approved):

qwrtln:
Hold on, there's a non-Roman numeral in the original that we need to fix first.

qwrtln:
Fixed in #977, all good.

PR #974, opened 2025-12-27
Fix unwanted git restore

qwrtln:


tools/build.sh
```diff
@@ -77,6 +77,15 @@ case "$(uname -s)" in
     ;;
 esac
 
+TEMP_DIR=$(mktemp -d)
+cleanup() {
+  if [[ -n "${SECTION_SEARCH}" ]]; then
```
DarthGandalf, line 82:
```suggestion
  if [[ -n "${TEMP_STRUCTURE}" ]]; then
```

Otherwise if exit on line 95 happens, it'll try to copy empty filename here?

PR #995, opened 2026-01-25
Use drop shadow image in encircle

qwrtln:
This fixes broken drop shadows on macOS preview and KDE okular by replacing vectors with PNG imgaes. Changes aren't visible at first glance.
en-11
en-16
en-18
en-22
en-23
en-24
en-25
en-28
en-34
en-46
en-57
en-59
en-65

### DIFF
--- a/translations/battlefield.tex/battlefield.tex.pot
+++ b/translations/battlefield.tex/battlefield.tex.pot
@@ -7,7 +7,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2026-01-12 21:10+0000\n"
+"POT-Creation-Date: 2026-01-28 22:23+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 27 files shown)


sections/units.tex
```diff
@@ -100,8 +100,8 @@
       \draw (-1.5, 0.6) node {\encirclesmall{5}};
       \draw (-1.5, -0.1) node {\encirclesmall{6}};
       \draw (-1.5, -0.8) node {\encirclesmall{7}};
-      \draw (0.3, -1) node {\encirclesmall{10}};
-      \draw (0.6, -2) node {\encirclesmall{11}};
+      \draw (0.3, -1) node {\encircle{10}};
```
DarthGandalf: this will be different size on the same picture then.

If the goal is to have 10+ always bigger size, this condition could be done inside the command

qwrtln: I wouldn't say this is always the intent. For big pictures, it's better to have the same (bigger) sizes everywhere. For smaller ones like this one with many circles, the big ones cover way too much. The change here affects only the drop shadow, not the circle size.

PR #1007, opened 2026-02-05
Spellscroll-Clarifications and Trading-section

DerSchlupf:
To do:

- [x] add link to trading section in spellscroll bullet point about selling spells
- [x] make more space between spellscroll and war machines
- [x] change wording, especially of first yellow bulletpoint: i'm still not happy with it, also with these triple \svg{epmpower}
- [x] discuss trading structure
- [x] maybe add a wiki-link to marketplace (https://en.homm3bg.wiki/events/marketplace/) as an example for "other trade options"

MISSING WIKI LINK

en-24

CLARIFICATIONS ABOUT SPELLSCROLLS:
**see here:** https://boardgamegeek.com/thread/3641907/two-questions-about-spell-scroll-cards 

LAYOUT SPELLSCROLLL CARDS:

en-26

LAYOUT TRADING:

en-50


sections/deckbuilding.tex
```diff
@@ -188,7 +188,7 @@ \subsection*{\pagetarget{Playerdecks}{Player Decks}}
 Spell cards\index{Spell card} have three possible primary effects.
 Using the topmost, basic version of the Spell usually has no additional costs.
 To access the other effects, you may \textbf{Empower} a Spell\index{Empowering Spells} by paying the indicated cost (3) to get a more powerful outcome (4).
-You may pay this cost by playing other cards for their Empower \svg{empower} effect (e.g. the Spell Power Statistic) before casting the Spell.
+You may pay this cost by playing other cards for their Empower \svg{empower} effect (e.g. the Statistic card \wikilink{statistics/power}{Spell Power}) before casting the Spell.
```
qwrtln:
```suggestion
You may pay this cost by playing other cards for their Empower \svg{empower} effect (e.g. the \wikilink{statistics/power}{Power} Statistic card) before casting the Spell.
```
The card name is "Power", though?

sections/deckbuilding.tex
```diff
@@ -380,7 +380,7 @@ \subsection*{\pagetarget{Schools of Magic}{Schools of Magic}}
 \begin{itemize}
     \item They don't count towards your Spell Limit.
     You can also play a Spell from your hand during this Combat Round.
-    \item You can only use the Spell's lowest Power \svg{empower} effect.
+    \item You may only use the Spell's lowest Power \svg{empower} effect, but even if it is 1\svg{empower} (e.g. \wikilink{spells/implosion}{Implosion}).
```
DarthGandalf: but?

sections/deckbuilding.tex
```diff
@@ -380,7 +380,7 @@ \subsection*{\pagetarget{Schools of Magic}{Schools of Magic}}
...
     You can't Empower the Spell with \svg{empower} from any source.
     % What, if the lowest Power is 1? Can it be still played or not?
```
qwrtln:
```suggestion
  \item You may only use the Spell's lowest Power \svg{empower} effect.
    You may not Empower the Spell with \svg{empower} from any source, even if the lowest effect costs more than 0~\svg{empower} (e.g. \wikilink{spells/implosion}{Implosion}).
```
This is difficult to phrase. How about this? First, the rule. Then, the limitation. Finally, the exception. I think it shouldn't say "1 spell power", because there are spells coming with the lowest effect costing 2.

DerSchlupf:
Hi, okay i've updatete the first post. Feel free to change wording, that it fits better. 

My idea with "but" was to say, that you can use that spell WITHOUT needing that one spellpower. you just play the card, regardless of their lowest effects cost.

sections/deckbuilding.tex
DerSchlupf: With this wording you could think that implosion wont work, cause you may not empower that spell but it required 1 spellpower. But the clarification is, that implosion works.

DerSchlupf: Or maybe something like that: 
\item You may use the Spell's lowest Power \svg{empower} effect, regardless of its Spell Power cost. But you may not choose a higher effect and you may not Empower the Spell with \svg{empower} from any source.

qwrtln: Better, yeah 👍🏻

DarthGandalf:
I find the answer "This should be together with trading" still ambiguous :(

It could mean as you understood, "in addition to any one of the other option", or it can be "only if you chose the Trading option, but not while removing cards and not while buying war machines"

DerSchlupf:
> I find the answer "This should be together with trading" still ambiguous :(
> 
> It could mean as you understood, "in addition to any one of the other option", or it can be "only if you chose the Trading option, but not while removing cards and not while buying war machines"

I wrote a follow up question in bgg.

sections/deckbuilding.tex
```diff
@@ -380,26 +380,22 @@ \subsection*{\pagetarget{Schools of Magic}{Schools of Magic}}
...
+    But you can't Empower the Spell with \svg{empower} from any source.
 \end{itemize}
 \end{expansion}
 \columnbreak
 \begin{expansion}{stronghold}
     \begin{itemize}
-    \item You can sell Spell Scrolls at a Trading Post for 2 \svg{gold} per each Spell on it.
-    %Is it a new option at a Trading Post or can i combine it with other actions in a single visit?
+    \item They can't be used to increase the \svg{empower} of another Spell.
+    \item You can sell Spell Scrolls at a Trading Post for 2 \svg{gold} per each Spell on it. This works in addition to the normal effect of Visiting the Trading Post.
```
qwrtln:
```suggestion
    \item You can sell Spell Scrolls at a Trading Post for 2 \svg{gold} per each Spell on it. This works in addition to the normal effect of visiting the Trading Post.
```

sections/deckbuilding.tex
```diff
@@ -380,26 +380,24 @@ \subsection*{\pagetarget{Schools of Magic}{Schools of Magic}}
 \begin{itemize}
     \item They don't count towards your Spell Limit.
     You can also play a Spell from your hand during this Combat Round.
-    \item You can only use the Spell's lowest Power \svg{empower} effect.
-    You can't Empower the Spell with \svg{empower} from any source.
-    % What, if the lowest Power is 1? Can it be still played or not?
-    \item They can't be used to increase the \svg{empower} of another Spell.
+    \item You may use the Spell's \textbf{lowest Power} \svg{empower} \textbf{effect}, without needing to pay its cost (e.g. in case of \wikilink{spells/implosion}{Implosion}).
```
DarthGandalf:
```suggestion
    \item You may use the Spell's \textbf{lowest Power} \svg{empower} \textbf{effect}, without needing to pay its cost (e.g. in case of \wikilink{spells/implosion}{Implosion} the lowest cost is 1).
```

DarthGandalf: Something is off with the wording here. But I can't produce good text for the example atm

qwrtln: I don't know, it's just an example, not a rule, so no need for that. @DerSchlupf?

DerSchlupf: I don't know too :D I would also say, we dont need it because of the wikilink, but if we have that space, it would be more clear.

qwrtln: Ok, added.

qwrtln (approved):

PR #1014, opened 2026-02-10
[DE] Update Playerturns

DerSchlupf:
Fuzzy Translation in this section are done. Unfortunatly i had to remove the negative moral art. The text in the right column was too long. I think the layout on page 17 could be improved, but i dont know how. Other than that its also ready for merge i think. 

LAYOUT
de-15
de-16
de-17


translations/player_turns.tex/de.po
```diff
@@ -366,16 +344,12 @@ msgstr ""
 "\\note{5}{Die Nekropolis \\svg{necro-yellow} Fraktion ignoriert sämtliche Moral-Effekte.\n"
 "  Sie kann niemals Moral erhalten oder verlieren.\n"
 "}\n"
-"\\columnbreak\n"
+"\\vspace{4em}\n"
```
qwrtln: Try this.
```suggestion
"\\vspace*{\fill}\n"
```

PR #1034, opened 2026-03-07
Move content to appropriate sections

### LINKED ISSUE #892: Move alternative combat boards to Variants?
These are optional rules after all

qwrtln:
Changes:
 * alternative combat boards: combat → variants
 * creature banks: expansion content → variants
 * empowered ability cards: renamed to empowered abilities and moved expansion content → deck-building
 * notes page: disabled
 
Page count reduced by 2 in printable, by 1 in digital (by putting the 3rd column of creature banks and alternative combat boards on a single page). See image below.

en-04

en-74

Fix #892


DerSchlupf:
Sry for beeing to late here. I'm not so happy with some of these changes. Let me explain:

1) I like having empowered ability cards in one place, but why splitting to diffenrent pages with other content in between. I suggest:
- move page 28 right after "statistics and abilties"
- change heading from "empowered abilities" (we already have this heading) into "empowered ability tokens" 
- make \subsection instead of \subheader
==> Then we had one double page with subheader "statistic and ability cards" and all rules about them.

2) I personally see a difference between "variants" and "expansion content": A variant is something, what changes or replaces current rules. "Expansion content" adds some completely new which wasnt there before. There are contents, which make both: They belong to an expansion, but replace current rules instead of introducing something new. I see following problems:
- Where do we draw the line? Why should we add creature banks there, but event cards not? And what about elemental Map tiles? Only, cause it fits in another chapter?  
- the combat boards, yes, they are "variants", but they are also expansion content. In the old version the variants chapter contains only core game variants. That had a certain elegance because all players could use it (even if they only had the core game. There were other chapters for expansion content. 

These are unfortunalty only problems. I havent a solution for now. Maybe we had to think/discuss about some general things: Which function do the different chapters have? What content do we want to see in each chapter?

PR #1037, opened 2026-03-10
whirlpool fix

DarthGandalf:
en-82
en-83


DerSchlupf (approved):

sections/all_map_locations.tex
```diff
@@ -772,8 +775,7 @@
     \expansionlabel{cove}
     \caption{{\raggedright \small Category: \textbf{Revisitable}}\\
       Move your Hero to another Whirlpool Token.
-      If there are multiple valid targets, roll an Attack Die to determine where your Hero goes.
-      Reroll any Die which shows the number of the Whirlpool your Hero is standing on.
+      If there are 3 Whirlpools, roll an Attack Die to determine where your Hero goes, and reroll any Die which shows the number of the Whirlpool your Hero is moving from.
```
qwrtln:
```suggestion
      If there are 3 Whirlpools, roll an Attack Die to determine where your Hero goes, and reroll any Die that shows the number of the Whirlpool your Hero is moving from.
```

sections/all_map_locations.tex
```diff
@@ -699,19 +699,23 @@
...
 \end{tikzpicture}}
 
 \newpage
 
-\pagetarget{Location Token}{\subheader{Location Tokens}}
-
-The following Fields are found on single-hex Location Tokens only, not on regular Map Tiles.
-The placement of the Location Tokens is described in \pagelink{Discover Location Tokens}{Map Elements}.
+\textbf{Note:} In the following Fields skip the move if you would be stepping onto allied Hero and don't have enough MP to move off it immediately.
+If the Field's effect placed your Hero on a Token on a still face down Tile, discover that Tile.
```
qwrtln:
```suggestion
\textbf{Note:} For the following Fields, skip that move if you would be stepping onto an allied Hero and do not have enough MP to move off it immediately.
If the Field's effect placed your Hero on a Token on a still face-down Tile, discover that Tile.
```

DarthGandalf: Why that instead of the?

qwrtln: Hmm, right, also on 2nd thought I think it should be "movement" rather than "move".

sections/all_map_locations.tex
```diff
@@ -699,19 +699,23 @@
...
+  \end{minipage}};
 \end{tikzpicture}}
 
 \newpage
 
-\pagetarget{Location Token}{\subheader{Location Tokens}}
-
-The following Fields are found on single-hex Location Tokens only, not on regular Map Tiles.
-The placement of the Location Tokens is described in \pagelink{Discover Location Tokens}{Map Elements}.
+\textbf{Note:} For the following Fields, skip that move if you would be stepping onto an allied Hero and do not have enough MP to move off it immediately.
```
qwrtln:
```suggestion
\textbf{Note:} For the following Fields, skip the movement if you would be stepping onto an allied Hero and do not have enough MP to move off it immediately.
```

qwrtln (approved):

PR #1054, opened 2026-03-28
update scans

DarthGandalf:
en-06
en-07
en-08
en-28
en-38
en-66
en-72
en-88


qwrtln (approved):
You can delete the old earth elemental images.

PR #1061, opened 2026-04-04
update attack and corrosion tokens

DarthGandalf:
en-09
en-89


qwrtln (approved):
Corrosion token looks good, I'd suggest tweaking the attack token to match the previous colors a little more, if you feel like it.

PR #1064, opened 2026-04-07
Don't limit faction cubes to hero in end of combat

DarthGandalf:
en-44

PR #1066, opened 2026-04-08
Fix victory conditions

DarthGandalf:
While reviewing what happens during town defeat for #1064, I noticed that we dropped a possibility for players to share victory, if after one player got all the enemy cubes, another player did the same in the same round.

Source: page 4 core mission book. 
изображение

en-11


sections/scenarios.tex
```diff
@@ -80,11 +80,13 @@ \subsection*{Tournament}
 Discard their Faction Cubes and Hero models from the Map.
 Treat the cards in their Deck as being removed from the game for the rest of the Scenario.
 If you are Eliminated, you may still participate in the game by controlling Neutral Units.\par
-\note{4}{If you Eliminate all enemy Factions, you immediately win the Scenario.}\par
+\note{4}{If you Eliminate all enemy Factions, you win the Scenario.}\par
```
qwrtln: What's wrong with immediate? If you eliminate everyone else, there's nothing left to do?

sections/scenarios.tex
```diff
@@ -80,11 +80,13 @@ \subsection*{Tournament}
...
 
-In \textbf{Clash} Scenarios with three or more players, collecting \textbf{a Faction Cube} from every enemy player immediately wins you the game.
+In \textbf{Clash} Scenarios with three or more players, collecting \textbf{a Faction Cube} from every enemy player wins you the Scenario.
```
qwrtln:
```suggestion
In \textbf{Clash} Scenarios with three or more players, when you collect \textbf{a Faction Cube} from every enemy player, you win the Scenario.
```
What about this draw situation in the other PR?

sections/scenarios.tex
```diff
@@ -80,11 +80,13 @@ \subsection*{Tournament}
...
 You gain such Faction Cubes by defeating Main Heroes or conquering enemy Towns.
 Other Scenario-specific rules may also modify the outcome of collecting Faction Cubes.
 
+Whenever a player achieves a Victory Condition, the game ends at the end of the Round, unless stated otherwise.
```
qwrtln: I would combine this with the first paragraph under this subheader. It feels out of place here.

sections/scenarios.tex
DarthGandalf: meaning?

qwrtln: Ah, sorry, it's in the description here. Well, you mentioned that you can have a draw, if the other player collects the same amount of faction cubes. Isn't it contradictory?

DarthGandalf: I don't see the contradiction

qwrtln: You win but not really, as somebody else can do that in the same round, so you don't win.

DarthGandalf: There could be several winners. At least that's how I'm reading the official book

sections/scenarios.tex
DarthGandalf: I'm not able to. For me moving this sentence there breaks the flow of text. How would you rephrase it?

qwrtln: WDYM? It's as simple as that.

Image

DarthGandalf: Then some conditions are listed before this sentence, some after. Alright

sections/scenarios.tex
```diff
@@ -69,6 +69,7 @@ \subsection*{Tournament}
 \pagetarget{End}{\subheader{Victory conditions}}
 All Scenarios have their victory conditions\index{Victory Conditions} described in the Mission Book.
 The most common victory conditions require defeating certain enemy Towns, Heroes, or Neutral Units; capturing the Grail; or controlling certain or most Locations over time.
+Whenever a player achieves a Victory Condition, the game ends at the end of the Round, unless stated otherwise.
```
qwrtln:
```suggestion
Whenever a player achieves required victory condition(s), the game ends at the end of the Round, unless stated otherwise.
```
1. Caps
2. I guess there could be more than one victory condition?

DarthGandalf: If the victory condition requires achieving several subconditions, that's still the condition. Adding (s) will make it more confusing wrt e.g. player elimination, which doesn't require other conditions to be satisfied.

qwrtln (approved):

qwrtln:
po4a commit

PR #1074, opened 2026-04-23
Faction cube town previous owner

DarthGandalf:
https://boardgamegeek.com/thread/3690983/capturing-enemy-town-from-another-enemy


DarthGandalf:
Unsure about the next sentence here. It may feel like original is also previous

qwrtln (approved):

PR #1086, opened 2026-05-06
Use English name for languages instead of transliterations

DarthGandalf:
From this list it becomes hard to find the correct name especially when such transliteration is not commonly used. And it becomes hard to understand what is what here


qwrtln:
Do it in the `build-release.yaml` file too, as this is a legacy script.

qwrtln (approved):

PR #1083, opened 2026-05-05
 Rule Mistake in Roundstructure Example ([DE] Translation included)

DerSchlupf:
This PR contains changes in the Original (only 1 Word) AND also the german translation of this one section. 

### Changes in the Original:

I've just edit one word in the round structure example: Instead of adding a Card to his DECK, Bob should add it to his HAND. 
image

### German Translation:

image
image

PR #1087, opened 2026-05-07
[CN] Chinese translation (ai_rules, all_map_locations)

qinsoon:
This PR adds Chinese translation for `ai_rules` and `all_map_locations`. It also adds fallback for Chinese for `LiberationSerif`.

ai_rules-01
ai_rules-02
ai_rules-03
ai_rules-04
all_map_locations-01
all_map_locations-02
all_map_locations-03
all_map_locations-04
all_map_locations-05
all_map_locations-06
all_map_locations-07
all_map_locations-08
all_map_locations-09
all_map_locations-10
all_map_locations-11
all_map_locations-12
all_map_locations-13


DarthGandalf:
Dragon utopia's description became center-aligned, was it intended?

qinsoon:
> Dragon utopia's description became center-aligned, was it intended?

Nice catch. The centering is weird.

The Chinese version for dragon utopia is structurally the same as the English one. The explanation from AI is that the Chinese translation is short enough and trigger latex caption’s `singlelinecheck`, and latex centers this particular caption while keeping others left-aligned. I added https://github.com/Heegu-sama/Homm3BG/pull/1087/commits/8085f2a0d23d6fb282778ff82f2afd17ac495846 to force left align specifically for this caption (dragon utopia) for the Chinese translation with `\\captionsetup{singlelinecheck=false, justification=raggedright}`.

all_map_locations-08-update

DerSchlupf:
If you wish — though it would be purely aesthetically — you can reduce the height of those note boxes like on first page of map locations, that it fit better the number of lines you have in Chinese translation.

You can keep this in mind for future note boxes in other sections

qinsoon:
> If you wish — though it would be purely aesthetically — you can reduce the height of those note boxes like on first page of map locations, that it fit better the number of lines you have in Chinese translation.
> 
> You can keep this in mind for future note boxes in other sections

Thanks for the note. One general question: does it become harder to maintain a translation when it diverges more from the English layout and formatting?

DerSchlupf:
Depends on language I would say. In German my common text length is higher than in English. I usually have to remove some art in some sections to keep the original page count. And sometimes (in case of really full-text-pages) I even need creative layout solutions to avoid shorting the text too much that the quality of rule explanations would not be reduced.

But if looking at your Chinese common textlength I wouldn't woory about difficult maintenance. It seems comfortable to keep page count 🙂

You even could add some new art on last page of AI section. There's pretty much space. Look in the asset/art folder which art is not in use at the moment and would fit. There should be some available.

DerSchlupf:
Here for example is an actual Problem I have in german: https://github.com/Heegu-sama/Homm3BG/pull/1083 😉 

Edit: just writing here I had an idea^^

qwrtln:
I'm not sure about Chinese, but in Latin scripts we tend to avoid having a lone word or (especially) a syllable in the last line of a paragraph, even though we can't always avoid it.

Screenshot From 2026-05-08 12-26-06

Please verify if your polygraphic standards follow a similar pattern.

qinsoon:
> I'm not sure about Chinese, but in Latin scripts we tend to avoid having a lone word or (especially) a syllable in the last line of a paragraph, even though we can't always avoid it.

Thanks for pointing this out. Yes, the same principle applies in Chinese typography as well -- short last lines or orphan characters at the end of a paragraph are usually avoided when possible. However, as in English and most other languages, this is more of an aesthetic refinement than a strict rule.

I didn’t realize that this level of detail was expected for the first translation pass; I had assumed it would be addressed later during a refinement/polishing pass before release.

I’ve now gone through and removed orphan cases as much as I reasonably can, and I’ve updated the comparison images in the post. Please take a look.

qwrtln:
> I’ve now gone through and removed orphan cases as much as I reasonably can, and I’ve updated the comparison images in the post. Please take a look.

It's not a strict requirement, just a recommendation. I wasn't sure if you were planning to make a second pass later. If so, those issues can be left out. Anyway, current render is an improvement over the first one 👍🏻 

For the next MR, I'd suggest including translated `language_metadata.tex`, so that any future renders get their final look.

qwrtln (approved):

PR #1088, opened 2026-05-08
Add a script to use coding agents to translate po files

qinsoon:
This is the script that I used for the Chinese translation workflow. The key idea is that to split po entries in a file, and give the agent one entry every time to translate. Compared to giving the agent a full file to translate, I find this approach more deterministic. If AI messes up or you run out of rate limit/token during work, you lose the progress of one entry rather than one file. This approach is also more cost efficient in terms of tokens. I completed the full translation using roughly two weeks of Codex rate limits, whereas letting agents freely work on entire files could easily burn through a week's quota without producing usable progress (in my experience).

The script currently uses codex CLI or claude CLI (which requires working codex/claude CLI with subscriptions). But it can be easily modified to support APIs.

Notes:
1. AI sometimes mishandle escaped LaTeX commands like `\\centering` well and may return incorrectly return `\centering` instead. In my experience, this happened in roughly 5–10% of cases. But mistakes like this are caught by `build.sh` so I manually fixed them. I tried to add this to the prompts, but AI still did not handle it correctly. A verification can be done in the script as well.


qwrtln:
We can have two workflows for the same thing, as this whole AI agentic thing is still experimental and exploratory. But anyway, a couple remarks:

1. The prompt seems largely duplicated. Wouldn't it be beneficial to have unified instructions loaded by both approaches, differing only by input data ingestion?
2. If you have your own glossary entries not covered by the existing ones, please expand them, so that every language can benefit.
3. The script doesn't seem to handle fuzzy entries. Are you planning to add this? This is a common pain point in this project, because the original gets updated frequently.

qinsoon:
>     1. The prompt seems largely duplicated. Wouldn't it be beneficial to have unified instructions loaded by both approaches, differing only by input data ingestion?

I extracted the translation rules from `.claude/commands/translate.md` into a separate file in the same folder. This new rule file is used by both `translate.md` and the new script I added.

>     2. If you have your own glossary entries not covered by the existing ones, please expand them, so that every language can benefit.

I added some entries to the glossary list.

>     3. The script doesn't seem to handle fuzzy entries. Are you planning to add this? This is a common pain point in this project, because the original gets updated frequently.

I added the handling for fuzzy entries, and quickly tested it. It seems working. The old entry, the old translation and the new entry are all provided to the agent to produce a new translation. The script will patch the new translation to the file. However, I haven't tested it much, and I don't have real cases to test how it works in practice. Please give it a try.

qinsoon:
I made this PR ready for review. Feel free to try it out (codex/claude CLI + subscription is needed -- if you use API, the script can modified to support API). Feel free to commit back to this PR if you encounter and fix any issue.

qwrtln (approved):
Okay, this Python script is a good starting point. Please upstream any changes that you make.

PR #1090, opened 2026-05-08
[CN] Chinese translation (language_metadata)

qinsoon:


qwrtln:
Not a must, but a suggestion: if there is nothing to translate, just copy the whole entry. Weblate and `po4a` will report translation progress more accurately. If you don't do this, the banner will never reach 100% ;-)

qwrtln (approved):

qinsoon:
> Not a must, but a suggestion: if there is nothing to translate, just copy the whole entry. Weblate and `po4a` will report translation progress more accurately. If you don't do this, the banner will never reach 100% ;-)

Thank you. I wasnt sure on that. I will update this PR.

qinsoon:
I updated the PR. It is ready to merge.

PR #1092, opened 2026-05-09
[DE] Update component-section

DerSchlupf:
de-06 (1)
de-07 (1)
de-08 (1)
de-09 (1)


DerSchlupf:
One Problem is left: Why does "Seechlacht_Erweiterung" not become a two-liner like the other labels on the right? It is in the whole chapter here.
image

DerSchlupf:
> One Problem is left: Why does "Seechlacht_Erweiterung" not become a two-liner like the other labels on the right? It is in the whole chapter here. image

Beside of that naval battles expansion label it should be now ready for merge. Pics in the first post. 
Would be nice if someone could fix that label-issue.

DarthGandalf:
This is effect of `\mbox` in that expansion in language metadata.

In English it was needed because "expansion" was split in a bad place, and half of the word went to next line.

DerSchlupf:
Pics in first post are up to date and this should be ready for merge. The Time Token link will be fixed in another PR (combat section). Any suggestions? Otherwise it would be nice to get approval.

qwrtln (approved):

PR #1093, opened 2026-05-09
[CN] Chinese translation (card_effect_keywords, combat)

qinsoon:
This PR adds Chinese translation for `card_effect_key_words` and `combat`.

card_effect_keywords-01
card_effect_keywords-02
card_effect_keywords-03
card_effect_keywords-04
combat-01
combat-02
combat-03-update
combat-04-update
combat-05-update
combat-06
combat-07
combat-08


qwrtln:
In general, don't repeat English `\colunmbreak`s, because they clearly don't work for Chinese.

qinsoon:
> In general, don't repeat English `\colunmbreak`s, because they clearly don't work for Chinese.

I updated p3-p5 in the combat section to handle columnbreaks. I also updated the images in the PR.
The columnbreak in p7 may need to be addressed as well, but that columnbreak is not included the translation file.

qwrtln (approved):
Art choice and alignment could be improved, but this is good enough for now 👍🏻

PR #1112, opened 2026-05-16
[PL] Heroes

qwrtln:
pl-20
pl-21


DarthGandalf:
what's liberation and size for?

qwrtln:
> what's liberation and size for?

To make the Roman numerals as similar as possible. Look at the "before" page 21. The difference is striking. Now they're almost identical.

DarthGandalf:
maybe make these numbers separate commands? or a command with number parameter

DarthGandalf:
Thanks. I'd name this something like `\herolevel` because encounter level doesn't have these wings

DarthGandalf (approved):

PR #1114, opened 2026-05-17
[CN] Index Chinese entries by pinyin

qinsoon:
This PR improves Chinese index generation by grouping entries according to the first romanized (pinyin) letter of each Chinese term. Previously, all Chinese index entries were grouped into a single “Chinese characters” section in main.

This PR also fixes a few mistakes for index entries.

cn-91-update


latexmkrc
```diff
@@ -15,7 +15,20 @@ if ($toggles) {
   $pre_tex_code = '\AtBeginDocument{'.$toggles.'}';
 }
 
+my %icu_locale_map = (
```
DarthGandalf, line 18: what's the point of this if ICU_LOCALE is set in .language_base.sh anyway?

qinsoon: `pdf-build.yaml` uses `latex-action` which directly builds the latex project using `latexmk`. That workflow does not use any script in the repo. So we need to set the local here again.

qwrtln: Then we should either:
1. Source the script in the `pre_compile` input of the latex action
2. Make this array latexmk-only

I'm leaning towards option 2, because this is the only place we need it.

qinsoon: I removed the locale table from `language_base.sh`. Now we only keep the table in `latexmk`.

qwrtln (approved):

PR #1118, opened 2026-05-19
[CN] Remove unneeded spaces in Chinese translations

qinsoon:
This PR removes most unnecessary spaces that appear within sentences and paragraphs in the Chinese translation. Unlike English and many other languages, Chinese generally does not use spaces as separators between words, so unexpected whitespace is visually distracting and typographically incorrect.

This PR systematically removes whitespace from the following cases:

* Spaces around LaTeX commands. Most of these spaces were preserved from the English source text due to the translation prompt instruction to “maintain all whitespace around LaTeX commands”.
* Newline characters (`\n`) that unintentionally introduce extra spaces between sentences. In English, LaTeX usually collapses adjacent whitespace into reasonable spacing, but in Chinese translations these newlines often result in visible unwanted spaces because Chinese text normally does not contain spaces between sentences.
* Explicit spaces introduced by `~`.

This PR also fixes a number of punctuation issues in the Chinese translation. Chinese punctuation is now used where appropriate (for example, `。` instead of `.`, and `（）` instead of `()`). Chinese punctuation typically occupies more visual space than its English counterpart, so `\wikilink` text in Chinese should generally exclude trailing punctuation. This ensures that the rendered brown circle appears immediately after the text, rather than after the punctuation mark.

There may also be some additional minor edits and consistency fixes included along the way.

The following is an example of white spaces that get removed in this PR. Due to the changes, there might be new orphan cases -- which I will fix as the last step.
cn-43-notes


DarthGandalf:
please resolve conflicts

DarthGandalf (approved):

Issue #1013, opened 2026-02-10
Big German Review


DerSchlupf:
Dear German Community,
Please post everything in this thread—any errors, inconsistencies, or ambiguities in the text or layout—that you notice while reading the German Fan Rules V.2.0 (experimental build). I will keep the list below up to date and fix the errors gradually.
_____________________________________
Liebe deutsche Community, 
schreibt in diesen Thread alles rein, was euch beim Lesen der deutschen Fan-Regeln V.2.0 (experimental build) an Fehlern/Unstimmigkeiten/Unklarheiten in Text oder Layout auffällt. Ich werde die Liste unten aktuell halten und nach und nach die Fehler beheben. 
____________________________________

**GERMAN CARD SCANS / FEHLENDE SCANS DEUTSCHER KARTEN**
(delivery of the new expansions required)
- [ ] empowerd necromancy (p.24)
- [ ] SpellscrollCard (p.27)
- [ ] neutral peasant (p. 37)
- [ ] bank dragon fly (p.36; p. 69)
- [ ] summon earth elemental (p.37)
- [ ] adventure tree of knowlegde (p.61)
- [ ] adventure cyclops stockpile (p. 61)
- [ ] offense and empowered offenese (p. 71)
- [ ] pandoras box: choose 2 ress OR MP pr Exp (p 73)
- [ ] manticore v2 (p. 73)
- [ ] few, pack earth elemantal (p.89)
- [ ] pit lord, demon (p. 89)

**CHECK ORIGINAL COMPONENT NAMES / ABGLEICH VON BEZEICHNUNGEN MIT DEN DEUTSCHEN ORIGINAL-REGELN**
(delivery of the new expansions required)
- [ ] entire keywords section: all token and unit names
- [ ] new locations names 
- [ ] creature bank locations, stack token
- [ ] emp ability token
- [ ] check some card titles: battlefield section (p. 58), tournanment (p. 53)
- [ ] names of new rule variants

**MISSING TRANSLATIONS AND TYPOS / FEHLENDE ÜBERSETZUNGEN ODER RECHTSCHREIBFEHLER:**

- [ ] p. 67 in ToC
- [x] S instead of P in creature bank combat (p. 69)
- [ ] "you have thrown" in moral actions 

**IMPROVABLE WORDING / VERBESSERUNGSWÜRDIGE FORMULIERUNGEN**
- [ ] ... 

[...]

PR #1129, opened 2026-05-26
[DE] Fine-Tuning Part 1

DerSchlupf:
Several small fixes are done. All of them are marked in the screenshots and are also listed in https://github.com/Heegu-sama/Homm3BG/issues/1013.

de-03
de-09
de-19
de-27
de-40
de-60
de-71

The red marking is the right one. Please ignore the yellow one:
de-73
de-90


DerSchlupf:
One problem still left: The text in specialty cards is centered. I dont know why. And a second question: how can i change the header as shown, but without getting a "two liner" in ToC? `\\` makes it a two-liner in ToC as far as i know. 
image

DerSchlupf:
And a second problem: Would be nice to have here a two-liner in the caption of minor artifacts as shown. But `\\` seems not to work. The place there should be enough. Any ideas?
image

DarthGandalf:
Done. The centering was because you missed newline after this paragraph, so the centering of the image below it accidentally applied to the text too

DerSchlupf:
Thanks for fixing the above issues!

i have a last one, before this PR should be ready for merge: one expansion bar is too long and goes into the bottom page border. thats a known issue, but at the last time you havent found a solution to fix it. maybe now? if not, then i would like to merge it without this fix. maybe someone found a solution later.
image

qwrtln (approved):

DerSchlupf:
po4a commit

PR #1130, opened 2026-05-26
learning stone affects main hero explicitly, so can give exp to

DarthGandalf:
secondary

en-74
ru-74


DerSchlupf:
Shouldn't we close this PR? According to "Astrologer Answers," the experience is lost—at least with the Treasure Die—but the same wording is used there, see screenshot. Or you like to keep it open, if any dev makes a newer clear announcement about that? 
Screenshot_20260619_090245_com_discord_MainActivity

PR #1135, opened 2026-05-30
[DE] german assets from first campaign

DerSchlupf:
To-Do-List:

- [ ] Check quality for assets
- [x] create "offens over wisdom" asset
- [x] scan empowered knowlegde
- [x] improve quality of adventure tree of knowlegde
- [x] fix "grottenlord" caption
- [x] make fireball asset bigger

LAYOUT
de-06
de-07
de-08
de-09
de-15
de-19
de-21
de-23
de-24
de-25
de-26
de-27
de-28
de-33
de-37
de-38
de-39
de-45
de-46
de-50
de-59
de-86
de-88


DerSchlupf:
This should be ready for review and merge. In rare cases in component section I used different cards, eg to reduce the number of scans or cause I haven't found the other campfire 😅 but that shouldn't be a problem I think.

DerSchlupf:
Only a small change: Artifact Labels with german words added
image

DarthGandalf:
Something happened on page 45 - shouldn't the image be wider? There's enough space for text above minotaur

DerSchlupf:
Sry for the late answer and thx for pointing it out. It is fixed now. Is there something more to do?
image

translations/assets/de/zombies_attack_griffins.png
DarthGandalf, line 1: Do you need both these files if only 1 is used?

DerSchlupf: Ah, no. I only want to keep the first one (instead replace it) until I know that the second one works. But now I will delete the unused one. 👍

qwrtln:
I suggest fixing drop shadows for sandro_empowering_lightning_bolt.png and skeletons_flipped.png, because they feel out of place now. You could just as easily regenerate them.

DerSchlupf:
The shadows are done and some other minor things. I have made updatet compared pages in the first post. Is it done now?

DarthGandalf:
po4a commit

DarthGandalf (approved):

PR #1146, opened 2026-06-14
[UA] Deckbuilding

bladetm13:
ua-22
ua-23
ua-24
ua-25
ua-26
ua-27
ua-28


qwrtln:
I'm not sure how you're building this, but apparently the texlive image used on CI makes the layout more sparse. Please download the PDF file from the workflow and inspect where it is broken.

bladetm13:
Got a problem with the index page here. Looks like I missed translating some terms, which caused the page to split into two. I will fix it as soon as I have some time.

bladetm13:
The problem is elsewhere. Some terms appear multiple times in the book. I translated one occurrence but not the others, so they started duplicating in the index. I can't translate the other sections separately - the same issue will pop up there. I'll check where the most duplicates are and probably add them translation here (I'll include a few sections in one PR, if you don't mind)

DarthGandalf:
It may be simpler to temporary allow index to consume more space than the page allows, then after the whole book is translated, probably that hack can be removed.

Resulting in something like this

ru-91

DarthGandalf:
That is done by this vspace

https://github.com/Heegu-sama/Homm3BG/blob/main/translations/index.tex/ru.po#L29

bladetm13:
Cool! I'll give it a try

qwrtln (approved):

PR #1149, opened 2026-06-16
Separate paragraph in AI siege for attacker and defender AIs

DarthGandalf:
ORIGINAL (English)
en-48

German unfuzzy:
image
image

PR #1153, opened 2026-06-19
Fix Chinese translations for AI siege

qinsoon:
This PR is based on https://github.com/Heegu-sama/Homm3BG/pull/1149


DarthGandalf:
Thanks, when creating the PR you could select that other branch as the target instead of main.

Or even push there directly

I've updated that branch with your commit

PR #1168, opened 2026-07-03
[RU] various small fixes, 2 non-RU fixes

DarthGandalf:
in sentence in intro - "." makes the passage about wiki too disconnected from ° symbol

add a link to expert effects which was never explained before this page


sections/round_structure.tex
```diff
@@ -9,7 +9,7 @@
 \begin{itemize}
   \item Flip any previously used Build, Population and/or Spell Book Tokens back to their active side.
   \item Flip any previously used Movement Point (MP) Tokens back to their active, green side.
-  \item Regain uses for Expert Effects \svg{expert}.
+  \item Regain uses for \pagelink{Ability}{Expert Effects \svg{expert}}.
```
qwrtln, line 12: Eh, I wonder if it's possible to make those SVGs "inherit" text color.

qwrtln (approved):

sections/round_structure.tex
DarthGandalf: I wonder the same. But it's not that important

PR #1171, opened 2026-07-05
Scenarios: secondary heroes also seem to give the cube

DarthGandalf:


DerSchlupf (approved):
Since only a single word was changed here, even I feel comfortable giving this my "Approval". It looks good to me and seems mergeable. 
However, this is also a test to see if my approval is actually sufficient to merge. ;)

PR #1175, opened 2026-07-06
Sell spell scroll as a whole, as alternative to other actions

DarthGandalf:
image

image

en-28
en-52

en-78


DerSchlupf:
I personally like this rule update much more then before and would be nice, if it becomes true. But actually we have another clarification from kamil and it is not so much old. Because of that I would wait with merging after we have a clear position, what's now the case.

DerSchlupf (approved):
I just read all changes and you got my approval. We should keep this changes in mind, until we got more clear statements from the devs.

qwrtln (approved):

PR #1183, opened 2026-07-18
Revise spell cards paragraph

qwrtln:
Efforts to make the "Spell Cards" paragraph simpler and more comprehensible:
en-18
en-26
en-29
en-30
en-33


DerSchlupf:
Some comments from my side:

1) I don't see a benefit to have the conflux expansion note there. It should be enough to have it in map elements. Cause there are so many reasons, which also increase spellpower. Wouldn't list only one of them. Otherwise we could also write: "Note: some Astrologers cards / permanent abilities / artifacts /... might also increase spells' spellpower."

2) I won't write "change midway which spell you're casting". AFAIK mardrax meant, that you can play empower effects before and you could change the spell you have "in mind". But as written players could think they could announce to play magic arrow, and later they change their decision, put magic arrow back to hand to play something else. That isn't hopefully possible? What about instead something like:
_Playing the cards together means to resolve the effects in any order, even to play "Empower" effects (e. g. Sorcery°) first, hoping to draw a better spell to cast._

3) in the current version I think we have not any detailed mage guild description. In both chapters towns and town actions we only have a short summary but maybe too short worded? I know your argument is that it is even written on the town board but I'm not sure if that building is important enough to get a bit more detailed explanation. I personally liked the previous one in spell chapter. Moving somewhere else is fine, but maybe it's too short now?

4) regarding layout on spell chapter: maybe switching spell card asset with the new note? Then the note would be directly near by the "play together" rule. But just an idea.

DarthGandalf:
> I don't see a benefit to have the conflux expansion note there.

The difference is that the abilities which do it, state explicitly that they do it. While the elemental tiles do not - they just have orange/blue colored background. My hope is that this mention causes the reader to pay attention later when he sees such tile.

> I won't write "change midway which spell you're casting"

This comes directly from [Alex's response](https://boardgamegeek.com/thread/3288802/article/44160894#44160894), where in that example magic arrow was put on the table first, before sorcery. But yeah, putting magic arrow back to hand was not intended consequence...

> mage guild description

no opinion

> regarding layout on spell chapter: maybe switching spell card asset with the new note

I had it like that, but switched this way because it seemed better when (3) and (5) mentioned above refer to the picture near it. Dunno.

sections/deckbuilding.tex
```diff
@@ -283,20 +288,23 @@ \subsection*{\pagetarget{Playerdecks}{Player Decks}}
   \end{itemize}
 \end{multicols*}
 
-Spells can be gained by Building the Mage Guild\index{Mage Guild}.
-Building it also enables you to use the Spell Book Token\index{Spell Book Token} for buying additional Spells.
-The Token cannot be used during the \textbf{same Round} when the Guild is constructed. Spells can also be \pagelink{Trading}{traded} in Alliance and Cooperative Scenarios.
+\vspace*{-0.5em}
+\note{6.5}{
+  Playing the cards together means that you can resolve the effects in any order, and even change midway which Spell you're casting, if e.g. you drew a better Spell using \wikilink{abilities/sorcery}{Sorcery.}
```
DerSchlupf:
```suggestion
  Playing the cards together means that you can resolve the effects in any order, and even play empower effects earlier, e.g. to hope drawing a better spell using \wikilink{abilities/sorcery}{Sorcery.}
```
I would highly recommmend to change the wording here in something like this. Please fix my grammar. :) But Alex answer is 2 years old and even mardrax said, that it was written down somewhat indistinctly. If it is permitted to switch spells that have already been announced to cast, that raises a whole lot of new questions. 

Sure, in practice your playgroup can allow it, when you ask them something like: "Sry people, i've said i want to play magic arrow, but i just draw lighting bold: Do you agree, i change the spell, maybe i use magic arrow to empower?" But that shouldnt be a rule in a rulebook.

sections/deckbuilding.tex
```diff
@@ -246,15 +246,19 @@ \subsection*{\pagetarget{Playerdecks}{Player Decks}}
...
+Using the topmost, basic version usually has no additional cost, but by paying the indicated cost \encircleinline{3} you can use a stronger effect \encircleinline{4}.
+You pay by playing other cards for their \textbf{Empower}\index{Empowering Spells} \svg{empower} effect (e.g. the \wikilink{statistics/power}{Power} Statistic card) together with the Spell.
+All the \svg{empower} played this way is consumed, though you may use a cheaper effect than the \svg{empower} you paid for.
+Each Spell card also has an alternative +1 \svg{empower} effect \encircleinline{5}.\par
 
-\bigskip
+See examples of Empowering on \hyperlink{example flagging mine}{page \pageref{example flagging mine}}.
+
+\begin{expansion}{conflux}
+  \textbf{Note:} \pagelink{Elemental Map Tiles}{Elemental Map Tiles} also increase Spells' \svg{empower}.
```
DerSchlupf:
```suggestion
  \textbf{Note:} \pagelink{Elemental Map Tiles}{Elemental Map Tiles} might also increase Spells' \svg{empower}.
```

What about adding "might", cause there are some requirements like main hero on that tile, fitting school. The Map Tiles don't necessarily provide spell power.

DerSchlupf:
I made two suggestions above. Other than that i'm still not fully happy to leave that mage guild text completely out, but if you both prefer it like it is now, i can live with it :) 

I havent really a good suggestion. but it feels now, that town actions and mage guild bullet point link to each other without giving really more details. As a reader, when I see a link, I expect to find more information there; instead, I am redirected back via another link.

xNeuroNx:
> > I don't see a benefit to have the conflux expansion note there.
> 
> The difference is that the abilities which do it, state explicitly that they do it. While the elemental tiles do not - they just have orange/blue colored background. My hope is that this mention causes the reader to pay attention later when he sees such tile.

It’s the same with every location, especially those marked with a “?”, or water, or underground tiles. It’s best if all these effects are listed in the “Map Effects” section or wherever that is. It would just be distracting if, while the basic rules are being explained, there’s suddenly “oh, and by the way, there’s also XYZ and ABC and...”. In my opinion, this directs the reader to a section they should read later, because you need to worry about map effects during gameplay, not while you’re learning the basic rules of cards.

sections/player_turns.tex
```diff
@@ -64,7 +64,8 @@
 \begin{itemize}
   \item [{\includegraphics[height=1.5\baselineskip, valign=c]{\images/build.png}}] Build Token\index{Build Token}, used to expand your \pagelink{Town}{Town}.
   \item [{\includegraphics[height=1.5\baselineskip, valign=c]{\images/population.png}}] Population Token\index{Population Token}, used to recruit and reinforce \pagelink{Units}{units} or to recruit \pagelink{Secondary}{a Secondary Hero}.
-  \item [{\includegraphics[height=1.5\baselineskip, valign=c]{\images/spells.png}}]Spell Book Token\index{Spell Book Token}, used to purchase \pagelink{spells}{Spells}. Requires the Mage Guild Town building to use.
+  \item [{\includegraphics[height=1.5\baselineskip, valign=c]{\images/spells.png}}]Spell Book Token\index{Spell Book Token}, used to purchase \pagelink{spells}{Spells}.
+    Requires the \pagelink{Town}{Mage Guild} Town building to be \textbf{built in a previous Round}.
```
DarthGandalf: Doesn't this mean that if it was built 2 turns ago, can't use the spell book anymore?

DarthGandalf: another corner case: if the mage guild was built from the start, then there's no previous turn on the first turn

qwrtln: "a pervious" means any previous round. "before this round" smells Russian - it doesn't have any anchor in English.
How would you handle "built from the start" case? I'm actually not sure if you're allow to buy spells on your first turn.

qwrtln: Oh, on your first turn with the mage guild already built you get the initial effect, so I suppose you can't spend the spell book token additionally?

DerSchlupf:
**Screenshots from Spellchapter below:** Layout must be optimized/ repaired.

Spell-Chapter:
image

DerSchlupf:
**Layout from Example page below:**
Layout have to be fixed, maybe we had to remove the resurrection art, i dont know.  Unfortunatly i wrote Lignthing Bol**T** wrong :/ 
The new asset should show the lighntning bold and the two other cards (sorcery; Magic arrow), which empower it. If we have more space left, we could add a sentence like the following, to have the fully sorcery advantages covered in this example:

_Had Alice drawn a non-magical card due to Sorcery, she would have had to cast Magic Arrow instead of Lightning Bolt and could have dealt only 2 damage instead of 4._

image

sections/map_elements.tex
```diff
@@ -138,24 +138,21 @@ \subsection*{Map Tile Anatomy}\index{Map Tile}
...
-
-\includegraphics[width=\linewidth]{\examples/sandro_empowering_lightning_bolt.png}
-
-\columnbreak
-\textit{The Combat lasted for only one Round, so Alice would not have been able to cast both Lightning Bolt and Haste, since players are limited to playing only one Spell card per Combat Round.}\par
+\textit{The Mine turns out to be guarded by Griffins, which have 4 HP \svg{health_points}.
+Alice's current hand consists of a Power card, a Magic Arrow, Sorcery, and a Town Portal.
+During the Combat, Alice plays Sorcery first hoping to draw an additional helpful card: She draws a Lightning Bolt, which she now decides to cast, and Empowers \svg{empower} it with the already played Sorcery's effect and Magic Arrows's alternative (bottom) effect, which makes the Lightning Bolt deal 4 damage \svg{damage}, killing the Griffins and winning the Combat.}
 
+\textit{The Combat lasted for only one Round, so Alice would not have been able to cast both Magic Arrow and the drawn Lightning Bold, since players are limited to playing only one Spell card per Combat Round.}\par
```
DerSchlupf:
```suggestion
\textit{The Combat lasted for only one Round, so Alice would not have been able to cast both Magic Arrow and the drawn Lightning Bolt, since players are limited to playing only one Spell card per Combat Round.}\par
```
:)

sections/player_turns.tex
DarthGandalf: the spell which you got at the start, are there from before the game started - from the moment the guild was build long time ago. Now during the first turn the guild is already built before this turn, so I can use spell book to buy more, why not

DerSchlupf (approved):
I just read all the changes and i am fine with all of them! But since it was highly debated PR, i suggest to wait for @Darth Gandalfs feedback/approval too before merging it.

DerSchlupf:
I dont know, why my outdatet suggestions were shown again there^^ 

But other question regarding german page count? Is there something to do now, or is it okay, if i solve that problem seperatly in an own german PR when i do the fuzzy?

EDIT: i could make the fuzzy in this PR too, if you want to wait with merge. I had time for it tomorrow evening

DarthGandalf:
> But other question regarding german page count? Is there something to do now, or is it okay, if i solve that problem seperatly in an own german PR when i do the fuzzy?

Better here. Otherwise to be able to merge this PR will need to add some hacks to German translation to make it fit anyway; you'd need to revert these hacks in your next PR

assets/gimp-files/sandro_empowering_lightning_bolt.xcf
DarthGandalf, line 1: let's remove this file?

DarthGandalf (approved):

assets/gimp-files/sandro_empowering_lightning_bolt.xcf
qwrtln: Let me do this in the PL unfuzzy PR.

PR #1184, opened 2026-07-18
Rework spells

DarthGandalf:
Alternative to #1183
en-26


DarthGandalf:
we decided to not introduce power pool here

